PR #8, opened 2018-08-03
Srs

fefrain:


common/app/srs.js
```diff
@@ -0,0 +1,72 @@
+import * as R from 'ramda'
```
minotogna: in common you need to user require('ramda'). es6 imports do not work in node

common/app/srs.js
```diff
@@ -0,0 +1,72 @@
...
+
+import codes from '@esri/proj-codes'
```
minotogna: same as above

webapp/commonComponents/dropdownComponent.js
```diff
@@ -85,16 +87,24 @@ class DropdownComponent extends React.Component {
...
 
     const contains = (a = '', b = '') => R.contains(R.toLower(a), R.toLower(b))
 
-    const filteredItems = R.trim(value).length > 0
-      ? items.filter(item => item.key
-        ? contains(value, item.key) || contains(value, item.value)
-        : contains(value, item)
+    const searchValue = R.trim(value)
+
+    let filteredItems
```
minotogna: here i would use ? : . i wouldn't use mutable objects with let

common/app/srs.js
fefrain: done

PR #14, opened 2018-08-10
Validation

fefrain:


server/auth/authApi.js
```diff
@@ -1,21 +1,26 @@
...
 
 const {sendOk} = require('../serverUtils/response')
 
 const {userPrefNames, getUserPrefSurveyId} = require('./../user/userPrefs')
 const {getSurveyById} = require('../survey/surveyRepository')
 const {deleteUserPref} = require('../user/userRepository')
+const {validateSurvey} = require('../survey/surveyValidator')
 
-const sendResponse = (res, user, survey = null) => res.json({user, survey})
+const sendResponse = (res, user, survey = null, surveyValidation = null) => res.json({user, survey, surveyValidation})
```
minotogna: i don't think you are passing the validation as parameter to the function. isn't it a property of the survey?

common/survey/nodeDef.js
```diff
@@ -42,6 +42,8 @@ const getNodeDefLabel = (nodeDef, lang) => R.pipe(
   R.prop(lang),
 )(nodeDef)
 
+const getNodeDefValidation = (nodeDef, field) => R.path(['validation', 'fields', field])(nodeDef)
```
minotogna: getNodeDefValidation would be a method that returns all validation object associated to the nodeDef. 
```javascript
const getNodeDefValidation = getProp('validation')
```
Then you might have another function to access filed specific validation, like 
```javascript
 const getNodeDefFieldValidation = (nodeDef, field) => R.pipe(
getNodeDefValidation,
R.path(['fields', field])
)(nodeDef)
```

server/nodeDef/nodeDefApi.js
```diff
@@ -40,12 +44,15 @@ module.exports.init = app => {
...
 
+    const {key, value} = body
+
     const nodeDefId = getRestParam(req, 'id')
 
     try {
       await updateNodeDefProp(nodeDefId, body)
-
-      sendOk(res)
+      const nodeDef = await fetchNodeDef(nodeDefId, true)
```
minotogna: you don' t need to fetch the nodeDef here. the updateNodeDefProp already returns the updated nodeDef.
this would be instead
```javascript
const nodeDef = await updateNodeDefProp(nodeDefId, body)
```

server/nodeDef/nodeDefValidator.js
```diff
@@ -0,0 +1,56 @@
...
+const {createError, validateRequired, assocValidation} = require('../serverUtils/validator')
+const {fetchNodeDefsBySurveyId} = require('./nodeDefRepository')
+const {getNodeDefProp, nodeDefType} = require('../../common/survey/nodeDef')
+
+const validateNodeDefName = async (nodeDef, newNode = true) => {
+  const requiredError = validateRequired('props.name', nodeDef)
+  if (requiredError)
+    return requiredError
+
+  const draft = true
```
minotogna: this can be inline.

server/survey/surveyApi.js
```diff
@@ -47,14 +47,16 @@ module.exports.init = app => {
...
     const surveyId = getRestParam(req, 'id')
 
+    const {key, value} = body
+
     try {
       await updateSurveyProp(surveyId, body)
-
-      sendOk(res)
+      const validation = await validateUpdateSurveyProp(surveyId, key, value)
+      res.json({status: 'ok', validation: validation})
```
minotogna: you don't need the status .
```javascript
res.json({validation})
```

server/survey/surveyRepository.js
```diff
@@ -57,8 +57,8 @@ const getSurveyById = async (surveyId, draft = false, client = db) =>
     def => dbTransformCallback(def, draft)
   )
 
-const getSurveyByName = async (surveyName, client = db) =>
-  await client.oneOrNone(
+const getSurveysByName = async (surveyName, client = db) =>
+  await client.many(
```
minotogna: can the query return 0 elements? If yes, you cannot use many, it will reject in that case.
You can use any.  http://vitaly-t.github.io/pg-promise/Database.html#many

temp/nodeDefValidation.js
```diff
@@ -0,0 +1,34 @@
+const nodeDefValidationType = {
```
minotogna: can this file be moved out of temp folder?

webapp/appUtils/asyncUtils.js
```diff
@@ -0,0 +1,9 @@
+export const runDelayed = (callback, key, delay = 500) => {
```
minotogna: can this function belong to reduxUtils? i think it is a redux utility..

webapp/survey/components/form/nodeDefEdit/commonProps.js
```diff
@@ -23,31 +26,58 @@ class CommonProps extends React.Component {
...
   render () {
     const {nodeDef, putNodeDefProp} = this.props
+    const {validation} = nodeDef
 
     return (
       <React.Fragment>
 
         <FormItem label={'name'}>
           <Input value={getNodeDefProp('name', '')(nodeDef)}
+                 validation={R.path(['fields', 'name'])(validation)}
```
minotogna: can you use getNodeDefFieldValidation here?

server/auth/authApi.js
fefrain: right...

common/survey/nodeDef.js
fefrain: Good suggestion!

server/nodeDef/nodeDefApi.js
fefrain: Ok

server/survey/surveyRepository.js
fefrain: Actually this should use "map" instead...

PR #33, opened 2018-11-05
Common survey refactor

minotogna:


webapp/appModules/home/components/surveysList.js
```diff
@@ -21,18 +19,18 @@ const SurveyListHeader = () => (
   </div>
 )
 
-const SurveyRow = ({survey, surveyInfo, setActiveSurvey}) => {
-  const surveyId = Survey.getSurveyId(survey)
+const SurveyRow = ({surveyInfoRow, surveyInfo, setActiveSurvey}) => {
```
fefrain: why is the parameter called "surveyInfoRow" and not just "surveyInfo"?

minotogna: because surveyInfo is the state object.

PR #58, opened 2018-11-26
Tables

sgiaccio:


webapp/appModules/surveyForm/components/items/items.scss
```diff
@@ -21,17 +21,31 @@
...
+  border-left: $itemsTableRowBorder;
 
-  button {
-    justify-self: center;
+  > :first-child {
+    margin-left: 1rem;
+  }
+
+  .align-right {
+    float: right;
```
minotogna: float is not necessary. it's better to use the flex justify

PR #71, opened 2018-12-04
fixed validation issues

fefrain:


webapp/appModules/surveyForm/taxonomyEdit/actions.js
```diff
@@ -97,6 +103,8 @@ export const uploadTaxonomyFile = (taxonomy, file) => async (dispatch, getState)
 
   dispatch(showAppJobMonitor(data.job, () => {
     //import complete
+    dispatch(fetchTaxonomy(surveyId, taxonomy.id))
+
     dispatch(reloadTaxa(taxonomy))
     //mark survey draft
```
minotogna: why do you first fetch the taxonomy and then reload taxa?

webapp/commonComponents/errorBadge.js
```diff
@@ -26,6 +29,7 @@ const ErrorBadge = ({validation}) => {
 
 ErrorBadge.defaultProps = {
   validation: null,
+  showInvalidLabel: true,
```
minotogna: can this just be showLabel?

webapp/appModules/surveyForm/taxonomyEdit/actions.js
fefrain: because we need to revalidate the taxonomy and show the just imported taxa. "fetchTaxonomy" does not fetch taxa.

webapp/commonComponents/errorBadge.js
fefrain: sure

webapp/appModules/surveyForm/taxonomyEdit/actions.js
minotogna: you could use Promise.all here, fetchTaxonomy and fetch taxa as well (with offset and limit params). let's try to minimize the number of dispatched actions.
here could you dispatch a taxonomyEditInit or Update with the taxonomy, the taxa, the offset and limit - see data/records

PR #82, opened 2018-12-13
Survey validations

fefrain:


common/survey/nodeDef.js
```diff
@@ -118,6 +118,9 @@ module.exports = {
 
   //advanced props
   getDefaultValues: SurveyUtils.getProp('defaultValues', []),
+  getCalculatedValues: SurveyUtils.getProp('calculatedValues', []),
+  getApplicable: SurveyUtils.getProp('applicable', []),
+  getExpressions: SurveyUtils.getProp('expressions', []),
```
minotogna: why does nodeDef have a property called expressions? did you mean validations?

server/nodeDef/nodeDefValidator.js
```diff
@@ -9,6 +9,14 @@ const {
 } = require('../../common/validation/validator')
 
 const NodeDef = require('../../common/survey/nodeDef')
+const NodeDefLayout = require('../../common/survey/nodeDefLayout')
+
+const NodeDefExpressionValidator = require('./nodeDefExpressionValidator')
+
+const errorKeys = {
```
minotogna: can this errorsKey be global?

server/nodeDef/nodeDefExpressionValidator.js
```diff
@@ -0,0 +1,43 @@
...
+  result.valid = R.pipe(
+    R.values,
+    R.none(v => !v.valid)
+  )(result.fields)
+
+  return result
+}
+
+module.exports = {
+  validateMultiple
```
minotogna: i would call the exposed method validated; then the internal method validate would become validateExpression. then the validateExpression would become validateJsepExpression. what do you think?

PR #85, opened 2018-12-18
Rdv node data types

sgiaccio:


common/dateUtils.js
```diff
@@ -0,0 +1,84 @@
...
+  differenceInWeeks,
+  differenceInDays,
+  differenceInHours,
+  format,
+
+  compareDesc,
+  isValid: fnsIsValid,
+} = require('date-fns')
+
+const addTrailingZeroes = (n, length) => {
```
minotogna: what about adding a stringUtils method, paddingLeft?
the method can be simpler like: return ("00000" + n).slice(-5)

common/dateUtils.js
```diff
@@ -0,0 +1,84 @@
...
+  return !!fnsIsValid(date)
+    && date.getUTCFullYear() === +year
+    && date.getUTCMonth() + 1 === +month
+    && date.getUTCDate() === +day
+}
+
+module.exports = {
+  getRelativeDate,
+  compareDatesDesc,
+  isValid,
```
minotogna: since this is a utils module, I would call this method isValidDate.
( it will also resolve the name conflict with date-fns.isValid )

common/dateUtils.js
```diff
@@ -0,0 +1,84 @@
...
+const isValid = (year, month, day) => {
+  if (!(year && month && day)) {
+    return false
+  }
+
+  const mm = addTrailingZeroes(month, 2)
+  const dd = addTrailingZeroes(day, 2)
+  const yyyy = addTrailingZeroes(year, 4)
+
+  const date = (new Date(`${yyyy}-${mm}-${dd}T00:00:00Z`))
```
minotogna: creating the date instance can be easier: new Date(yyyy, mm, dd) - parameters are integer; 
you won't need the addTrailingZeroes in this case

server/surveyRdb/schemaRdb/dataColProps.js
```diff
@@ -54,6 +60,22 @@ const props = {
     [colTypeProcessor]: () => () => sqlTypes.decimal,
   },
 
+  [nodeDefType.date]: {
+    [colTypeProcessor]: () => () => sqlTypes.date,
+    [colValueProcessor]: (surveyInfo, nodeDefCol, nodeCol) => {
+      const [day, month, year] = Node.getNodeValue(nodeCol, '').split('/').map(trim)
```
minotogna: are the values in the db saved as dd/mm/yyyy ? I would have saved them as yyyy-mm-dd

server/surveyRdb/schemaRdb/dataColProps.js
```diff
@@ -54,6 +60,22 @@ const props = {
...
+      return () => isValid(year, month, day) ? `${year}-${month}-${day}` : null
+    }
+  },
+
+  [nodeDefType.time]: {
+    [colTypeProcessor]: () => () => sqlTypes.time,
+    [colValueProcessor]: (surveyInfo, nodeDefCol, nodeCol) => {
+      const [hour, minute] = Node.getNodeValue(nodeCol, '').split(':').map(trim)
+      return () => hour !== '' && minute !== '' ? `${hour}:${minute}:00` : null
```
minotogna: you can use stringUtils.isNotBlank and avoid the .map(trim)

common/dateUtils.js
sgiaccio: This is the only reliable way I found to verify the validity. new Date(2012, 0, 1) gives 2011-12-31T23:00:00.000Z. new Date('2012-01-01T00:00:00Z') properly gives 2012-01-01T00:00:00.000Z. moment.js would probably solve these problems

server/surveyRdb/schemaRdb/dataColProps.js
sgiaccio: Ok

common/dateUtils.js
sgiaccio: Ok, can do paddingLeft(str, paddingChar, n)

common/dateUtils.js
minotogna: have you checked with date-fns?

minotogna: I just tested this in the browser: 
new Date(2012, 0, 1)
Sun Jan 01 2012 00:00:00 GMT+0100 (Central European Standard Time)

can you specify how you managed to get that date?

PR #87, opened 2018-12-19
Expressions validation

fefrain:


common/exprParser/exprParser.js
```diff
@@ -31,9 +31,12 @@ const memberExpression = async (expr, ctx) => {
...
-  return isString(objectRes)
-    ? eval(`${objectRes}.${propertyRes}`)
-    : R.prop(propertyRes, objectRes)
+  if (isString(objectRes))
+    return eval(`${objectRes}.${propertyRes}`)
+  else if (R.has(propertyRes, objectRes))
+    return R.prop(propertyRes, objectRes)
+  else
+    throw new Error(`Invalid property ${propertyRes}`)
 }
```
minotogna: undefined is returned in javascript if you access a property that doesn't exist in a object . why should our expression parser throw an error?

fefrain: it prevents to define expressions like "this.invaildProperty" or this.parent().unexistingProperty"

PR #100, opened 2019-01-18
Record permissions

sgiaccio:


common/survey/_internal/surveyDefaults.js
```diff
@@ -19,6 +19,8 @@ const getDefaultAuthGroups = (lang) => [
     ],
     labels: {[lang]: 'Survey administrators'},
     descriptions: {[lang]: `Full rights`},
+
+    dataSteps: {'1': 'all', '2': 'all', '3': 'all'},
```
minotogna: please, use keys, not strings

server/record/recordApi.js
```diff
@@ -8,10 +8,15 @@ const Node = require('../../common/record/node')
 const RecordManager = require('./recordManager')
 const FileManager = require('../file/fileManager')
```
minotogna: are delete Node and delete records managed?

server/authGroup/authGroupRepository.js
```diff
@@ -8,15 +8,16 @@ const dbTransformCallback = camelize
 
 const insertGroup = async (authGroup, surveyId, client = db) =>
   await client.one(`
-    INSERT INTO auth_group (name, survey_id, permissions, labels, descriptions)
-    VALUES ($1, $2, $3, $4, $5)
+    INSERT INTO auth_group (name, survey_id, permissions, labels, descriptions, data_steps)
```
minotogna: should the column called record_steps?

PR #106, opened 2019-01-23
Client side record permissions

sgiaccio:


webapp/appModules/data/records/components/recordView.js
```diff
@@ -45,19 +44,22 @@ class RecordView extends React.Component {
   }
 
   render () {
-    const {recordLoaded, preview} = this.props
+    const {recordLoaded, preview, user, record} = this.props
 
     return recordLoaded
-      ? <SurveyFormView draft={preview} preview={preview} edit={false} entry={true}/>
+      ? <SurveyFormView draft={preview} preview={preview} edit={false} entry={true} canEditRecord={canEditRecord(user, record)}/>
```
fefrain: do not pass user and record to props, calculate "canEditRecord" prop in mapStateToProps

webapp/appModules/data/records/components/recordsView.js
```diff
@@ -10,6 +10,7 @@ import TablePaginator from '../../../../commonComponents/table/tablePaginator'
 
 import Survey from '../../../../../common/survey/survey'
 import NodeDef from '../../../../../common/survey/nodeDef'
+import { canEditRecord } from '../../../../../common/auth/authManager'
```
fefrain: our convention is to import the entire object and call function from it. e.g. AuthManager.canEditRecord

webapp/appModules/data/records/components/recordsView.js
```diff
@@ -34,14 +36,20 @@ const RecordRow = ({idx, offset, record, style, nodeDefKeys}) => (
     <div>{getRelativeDate(record.dateModified)}</div>
     <div>{record.ownerName}</div>
     <div>
-      <Link to={appModuleUri(dataModules.record) + record.uuid} className="btn btn-s btn-of-light-xs">
-        <span className="icon icon-pencil2 icon-12px"></span>
-      </Link>
+      {
+        canEdit
```
fefrain: if the change is only in the icon, make only the icon class dynamic

webapp/appModules/surveyForm/nodeDefs/components/nodeDefTableBody.js
```diff
@@ -83,7 +83,11 @@ class NodeDefMultipleTableBody extends React.Component {
           <span className="values-summary">{this.state.nodeValues}</span>
           <button className="btn-s btn-of-light-xs"
                   onClick={() => this.toggleEditDialogOpen(true)}>
-            <span className="icon icon-pencil2 icon-12px"/>
+            {
+              this.props.canEditRecord
+                ? <span className="icon icon-pencil2 icon-12px" />
```
fefrain: same as above, make only class name dynamic

webapp/appModules/data/records/components/recordView.js
minotogna: user and   record are still in mapStateToProps, do not pass them to props,

PR #131, opened 2019-02-27
Data vis csv export

fefrain:


server/surveyRdb/surveyRdbManager.js
```diff
@@ -37,6 +40,34 @@ const countTable = async (surveyId, tableName, filter, client = db) =>
...
 
+const exportTableToCSV = async (surveyId, tableName, cols, filter, output) => {
+  const csvStream = fastcsv.createWriteStream({ headers: true })
+  csvStream.pipe(output)
+
+  // 1. write headers
+  csvStream.write(cols)
+
+  let offset = 0
+  const limit = 10
```
minotogna: Limit at least 100 for csv export?

server/serverUtils/response.js
```diff
@@ -29,16 +33,24 @@ const sendErr = (res, err) => {
...
-  res.setHeader('Content-Disposition', `attachment; filename=${File.getName(file)}`)
-  res.setHeader('Content-Length', File.getSize(file))
-  // res.set('Content-Type', 'text/csv')
-
+  sendFilePrepare(res, File.getName(file), File.getSize(file))
   res.write(file.content, 'binary')
   res.end(null, 'binary')
 }
 
+const sendFilePrepare = (res, fileName, fileSize = null, contentType = null) => {
```
minotogna: I would call this function setFileOutput or similar?

server/surveyRdb/surveyRdbManager.js
```diff
@@ -37,6 +40,34 @@ const countTable = async (surveyId, tableName, filter, client = db) =>
 const queryRootTableByRecordKeys = async (survey, recordUuid, client = db) =>
   await TableViewQuery.queryRootTableByRecordKeys(survey, recordUuid, client)
 
+const exportTableToCSV = async (surveyId, tableName, cols, filter, output) => {
+  const csvStream = fastcsv.createWriteStream({ headers: true })
```
minotogna: I would create a separate module TableViewExport (see TableViewQuery)

server/surveyRdb/surveyRdbManager.js
fefrain: sure, the limit to 10 is simply a mistake... we can limit the query even to 500 rows, it won't be too slow and it won't occupy too much memory

server/serverUtils/response.js
fefrain: it's a function that modify a response, setting the content type and disposition... it's more something like "prepare to file write"

server/surveyRdb/surveyRdbManager.js
fefrain: I agree...

server/serverUtils/response.js
minotogna: Yes you're setting whatever the response needs to output a file. I was commenting about what is easier to understand when you read ```Response.sendFilePrepare(...)``` or ```Response.setContentTypeFile(...)``` or better name... What do you think?

PR #133, opened 2019-03-05
Collect survey import

fefrain:


server/survey/collectImport/jobs/surveyCreatorJob.js
```diff
@@ -0,0 +1,37 @@
+const R = require('ramda')
+
+const Survey = require('../../../../common/survey/survey')
+const SurveyInfo = require('../../../../common/survey/_internal/surveyInfo')
```
minotogna: this is internal to survey, it cannot be exported here. export the SurveyInfo.keys from Survey as infoKeys

server/job/job.js
```diff
@@ -34,6 +34,7 @@ class Job {
     this.processed = 0
     this.result = {}
     this.errors = {}
+    this.context = {} // object shared among nested jobs
```
minotogna: job starts to have  a lot of properties.. should user, surveyId, survey, and/or others be props of context?

server/survey/collectImport/jobs/collectIdmlParseUtils.js
```diff
@@ -0,0 +1,40 @@
+const R = require('ramda')
```
minotogna: I would call this file collectIdmlParser.

server/survey/collectImport/jobs/surveySchemaReaderJob.js
```diff
@@ -0,0 +1,38 @@
...
+    const options = {
+      attrNodeName: '_attr',
+      attributeNamePrefix: '',
+      textNodeName: '_text',
+      ignoreAttributes: false,
+      format: false,
+      indentBy: '  ',
+    }
+    const idmlTraversalObj = parser.getTraversalObj(idmlXml, options)
+    const idmlJsonObj = parser.convertToJson(idmlTraversalObj, options)
```
minotogna: I would add all this parsing code to the object collectIdmlParser

server/survey/collectImport/jobs/surveySchemaReaderJob.js
```diff
@@ -0,0 +1,38 @@
...
+
+    this.setContext({zipFile, surveySource: idmlJsonObj.survey})
```
minotogna: i would call this context prop collectSurvey (to add the collect prefix to distinguish this prop within other jobs)

server/system/appCluster.js
```diff
@@ -23,7 +23,9 @@ module.exports = async () => {
   app.use(fileUpload({
     //limit upload to 10MB
     limits: {fileSize: 10 * 1024 * 1024},
-    abortOnLimit: true
+    abortOnLimit: true,
+    useTempFiles : true,
+    tempFileDir : '/tmp/'
```
minotogna: why do you need this? (i'm asking because it must be checked when and where we'll deploy the app)

server/taxonomy/taxonomyImportHelper.js
```diff
@@ -0,0 +1,63 @@
+const R = require('ramda')
```
minotogna: this file should probably be a TaxonomyImportService?

webapp/survey/actions.js
```diff
@@ -5,14 +5,15 @@ import { getUser } from '../app/appState'
 import { userPrefNames } from '../../common/user/userPrefs'
 import { appUserPrefUpdate } from '../app/actions'
 import { showAppJobMonitor } from '../appModules/appView/components/job/actions'
+import Node from '../../common/record/node'
```
minotogna: this is not used

server/survey/collectImport/jobs/surveyCreatorJob.js
fefrain: rightissimo!

server/survey/collectImport/jobs/collectIdmlParseUtils.js
fefrain: in that case it should be a class, not just a set of helper methods

server/survey/collectImport/jobs/surveySchemaReaderJob.js
fefrain: it doesn't do too much, it's would be just a generic method that parses XML into JSON... it's not related to the Collect IDML

server/system/appCluster.js
fefrain: it avoids the serialization of the uploaded files when starting jobs

server/taxonomy/taxonomyImportHelper.js
fefrain: no, it's used internally by the TaxonomyImportJob and CollectTaxonomiesImportJob

server/survey/collectImport/jobs/collectIdmlParseUtils.js
minotogna: no but why? just for three methods?

server/survey/collectImport/jobs/surveySchemaReaderJob.js
minotogna: yes i know, it was more about to isolate this logic and all dependencies into the collectIdmlParser (which could be moved at this level in the directory?)

server/system/appCluster.js
minotogna: would this work on an aws (i haven't checked...) or. . let's leave a TODO?

PR #136, opened 2019-03-07
Data sorting

sgiaccio:


server/surveyRdb/dbActions/tableViewQuery.js
```diff
@@ -68,12 +69,12 @@ const queryRootTableByRecordKeys = async (survey, recordUuid, client) => {
     keyNodes.map(
       async node => {
         const colValue = await getColValue(node)
-        return `${getColName(node)} ${colValue === null ? ' IS NULL' : " = '" + colValue + "'"}`
+        return `${getColName(node)} ${colValue === null ? ' IS NULL' : ' = \'' + colValue + '\''}`
```
minotogna: is this escaping needed ?

webapp/commonComponents/form/input.js
```diff
@@ -24,21 +24,24 @@ export const Input = React.forwardRef((props, ref) => {
   } = props
 
   return (
-    <TooltipError messages={validation.errors}>
+    <Tooltip
+      messages={validation.errors}
+      type={`${validation.errors ? 'error' : ''}`}
+      style={{ display: 'grid', width: '100%', height: '100%' }}>
```
minotogna: why would you need this inline style?

webapp/commonComponents/tooltip.js
```diff
@@ -3,42 +3,53 @@ import './tooltip.scss'
...
+    </div>
+  }
 }
 
-export const TooltipError = (props) => (
-  <Tooltip {...props}
-           type="error"/>
-)
+Tooltip.defaultProps = {
+  message: null,
```
minotogna: I think it's messages... could you also add the other props?

server/surveyRdb/dbActions/tableViewQuery.js
sgiaccio: It's not... Must have been the editor...

webapp/commonComponents/form/input.js
sgiaccio: Yes it doesn't look good, I'll find a better solution

webapp/commonComponents/tooltip.js
sgiaccio: Will do

minotogna:
Issues:
When I set the filter, then i navigate to another page, when i come back to the dataQuery on mouse over the filter button the sort is not displayed

PR #156, opened 2019-04-29
Sort and query management

sgiaccio:


server/modules/surveyRdb/persistence/dbActions/tableViewQuery.js
```diff
@@ -9,33 +9,39 @@ const Node = require('../../../../../common/record/node')
 const SchemaRdb = require('../../../../../common/surveyRdb/schemaRdb')
 const NodeDefTable = require('../../../../../common/surveyRdb/nodeDefTable')
 
+const SqlUtils = require('../../../../../common/exprParser/helpers/sqlUtils')
```
minotogna: exprParser/helpers have to be consider as internal helper modules. you can export the sqlUtils methods from exprParser/expression

common/exprParser/helpers/sqlUtils.js
```diff
@@ -0,0 +1,54 @@
...
+    return 'AND'
+  case '===':
+    return '='
+  case '!==':
+    return '!='
+  default:
+    throw new Error(`Unknown operator: ${op}`)
+  }
+}
+
```
minotogna: I wouldn't use a switch. but a simple key value object. e.g. {'||':'OR' , '&&' : 'AND' }

common/exprParser/helpers/sqlUtils.js
```diff
@@ -0,0 +1,54 @@
...
+const binaryToString = (node, params) => {
+  const left = toString(node.left, params)
+  const right = toString(node.right, left.params)
+
+  return {
+    str: `${left.str} ${getSqlOperator(node.operator)} ${right.str}`,
```
minotogna: and here use R.prop(node.operator, sqlOperators)

server/modules/surveyRdb/api/surveyRdbApi.js
```diff
@@ -40,7 +41,7 @@ module.exports.init = app => {
       const surveyId = Request.getRequiredParam(req, 'surveyId')
       const tableName = Request.getRequiredParam(req, 'tableName')
       const cols = Request.getJsonParam(req, 'cols', [])
-      const filter = Request.getRestParam(req, 'filter', '')
+      const filter = Request.getJsonParam(req, 'filter', null)
```
minotogna: defaultValue is by default null. it can be omitted here

server/modules/surveyRdb/persistence/dbActions/tableViewQuery.js
```diff
@@ -9,33 +9,39 @@ const Node = require('../../../../../common/record/node')
...
     `,
-    []
+    { ...filterParams, ...colParams, schemaName, tableName, limit, offset }
   )
 }
 
-const runCount = async (surveyId, tableName, filter = '', client) => {
+const runCount = async (surveyId, tableName, filterExpr, client) => {
   const schemaName = SchemaRdb.getName(surveyId)
+  const { str: filterQuery, params: filterParams } = filterExpr ? SqlUtils.getWherePerparedStatement(filterExpr) : {}
```
minotogna: here as above, don't access directly to sqlUtils method

server/modules/surveyRdb/persistence/dbActions/tableViewQuery.js
```diff
@@ -9,33 +9,39 @@ const Node = require('../../../../../common/record/node')
...
 
   return await client.one(`
     SELECT count(*) 
     FROM ${schemaName}.${tableName}
```
minotogna: should you use here as well FROM $/schemaName:name/.$/tableName:name/ as in runSelect method?

webapp/loggedin/modules/data/dataVis/dataQuery/components/tableHeader.js
```diff
@@ -57,14 +58,14 @@ class TableHeader extends React.Component {
...
-    const csvDownloadLink = `/api/surveyRdb/${surveyId}/${tableName}/export?filter=${filter}&sort=${DataSort.serialize(sort)}&cols=${JSON.stringify(colNames)}`
+    const csvDownloadLink = `/api/surveyRdb/${surveyId}/${tableName}/export?filter=${JSON.stringify(filter)}&sort=${DataSort.serialize(sort)}&cols=${JSON.stringify(colNames)}`
     const sortMsg = DataSort.getViewExpr(sort)
 
     return (
       <div className="table__header">
 
         <div className="data-operations">
-          <Tooltip messages={filter && [filter]}>
+          <Tooltip messages={R.isNil(filter) ? [] : [Expression.toString(filter, Expression.modes.sql)]}>
```
minotogna: can you still use 
```messages = {filter && [Expression.toString(filter, Expression.modes.sql)]}``` ?

File: webapp/loggedin/modules/data/dataVis/dataQuery/actions.js [hunk 8]
```diff
@@ -125,7 +123,11 @@ export const initTableData = (queryFilter = null, querySort = null, editModePara
       const { offset, limit } = defaults
 
       const [countResp, dataResp] = await Promise.all([
-        axios.get(`/api/surveyRdb/${surveyId}/${tableName}/query/count?filter=${filter}`),
+        axios.post(
+          `/api/surveyRdb/${surveyId}/${tableName}/query/count`, {
+            filter: filter ? JSON.stringify(filter) : null,
```
minotogna: can you use 
```filter: filter && JSON.stringify(filter)``` ?

File: webapp/loggedin/modules/data/dataVis/dataQuery/components/tableHeader.js [hunk 9]
```diff
@@ -74,10 +75,10 @@ class TableHeader extends React.Component {
             showExpressionEditor &&
             <ExpressionEditorPopup
               nodeDefUuidContext={nodeDefUuidContext}
-              query={filter}
+              expr={filter}
               mode={Expression.modes.sql}
-              onChange={query => {
-                updateTableFilter(query)
+              onChange={(_, queryExpr) => {
```
minotogna: isn't this the AST tree expr? I would call this expr, it confuses me queryExpr

webapp/commonComponents/expression/expressionEditorPopup.js
```diff
@@ -93,6 +93,7 @@ class ExpressionEditorPopup extends React.Component {
 
 ExpressionEditorPopup.defaultProps = {
   query: '',
+  expr: null,
```
fefrain: please add some comments on the parameters

webapp/loggedin/modules/data/dataVis/dataQuery/components/tableHeader.js
sgiaccio: Not anymore, we have to check that filter is not null in tooltip.js to do so

common/exprParser/helpers/sqlUtils.js
sgiaccio: Kept it as a function to throw an exception if the operator is unknown

common/exprParser/helpers/sqlUtils.js
```diff
@@ -0,0 +1,54 @@
+const { types } = require('./types')
+
+const getSqlOperator = op => {
+  const sqlOperator = {
```
fefrain: I would move this object outside, in a constant like sqlOperatorByJsOperator!

webapp/loggedin/modules/data/dataVis/dataQuery/components/tableHeader.js
fefrain: @sgiaccio is right, "filter" now can be null

webapp/commonComponents/expression/expressionEditorPopup.js
```diff
@@ -74,12 +75,12 @@ class ExpressionEditorPopup extends React.Component {
         <div className="expression-editor-popup__footer">
           <button className="btn btn-xs btn-of"
                   onClick={() => onChange('')}
-                  aria-disabled={R.isEmpty(query)}>
-            <span className="icon icon-undo2 icon-16px"/> Reset
+                  aria-disabled={!Expression.toString(expr)}>
```
minotogna: why can't you use query here?

sgiaccio: Ok to the getSqlOperator but I think it's safer to throw an exception
Query can be null also if expr is null as depending on the parent, either query or expr are passed

webapp/loggedin/modules/data/dataVis/dataQuery/components/tableHeader.js
minotogna: but if filter is null, if you do ```messages = {filter && [Expression.toString(filter, Expression.modes.sql)]}```, wouldn't it return null?

sgiaccio: Yes we can do that, then we need to add an additional control on Tooltip

minotogna: that would be better I think

webapp/commonComponents/expression/expressionEditorPopup.js
minotogna: that was not my question.. this answers probably refers to another question

PR #163, opened 2019-04-30
add log module

fefrain:


server/modules/log/log.js
```diff
@@ -0,0 +1,24 @@
...
+    }
+  }
+})
+
+const getLogger = () => log4js.getLogger('arena')
+
+module.exports = {
+  isDebugEnabled: () => getLogger().isDebugEnabled(),
+
+  debug: message => getLogger().debug(message),
```
minotogna: you need to check with isDebugEnabled, otherwise the operation might be consuming unnecessary resources

server/modules/log/log.js
```diff
@@ -0,0 +1,24 @@
...
+  },
+  categories: {
+    default: {
+      appenders: ['console'], level: 'debug'
+      //appenders: ['file'], level: 'error'
+    }
+  }
+})
+
+const getLogger = () => log4js.getLogger('arena')
```
minotogna: this can be ```logger = log4js.getLogger('arena')```

server/modules/log/log.js
fefrain: no, you need to check isDebugEnabled if the generation of the message you have to log is time consuming; if the level is higher than debug, the message won't be processed by any appenders (https://stritti.github.io/log4js/docu/users-guide.html#performance)

server/modules/log/log.js
fefrain: good catch

server/modules/log/log.js
minotogna: My question is: if our logger.dubug is wrapped into our log module, how can you know if 
 the generation of the message you have to log is time consuming??

In the page you sent me, I cannot find what you wrote, but only:

```To improve the performance of your software, it is a good idea to surround your logging statements with appropriate isXXXEnabled() methods to evaluate before logging if the required logging level is enabled:
```if( myLogger.isDebugEnabled()) {
	myLogger.debug(...);
  }```
If debug is disabled, it is much faster to evaluate only if it is disabled than constructing a possible complex message before finding debug level disabled. But this makes only sense when the messages are a little bit more complex than just log a simple string. Otherwise the sources are blown up too much with condition statements which is not much readable.

fefrain: As even this guy says: "But this makes only 
sense when the messages are a little bit more complex than just log a simple string. 
Otherwise the sources are blown up too much with condition statements which is not 
much readable."

it makes sense in cases where you have something like log.debug(`${formatDate()... ${strangeHeavyCalculation()}...`) in this case the string to log can be complex to generate.
Anyway, since we are wrapping the logger, we can always check if debug is enabled before sending the message to it, but it's a bit useless because we still need to perform it before calling "debug" in the case above.

minotogna: alright, then, you need to expose isLevelXXXEnabled.

PR #165, opened 2019-04-30
Survey form state refactor

minotogna:


server/job/job.js
```diff
@@ -55,6 +57,8 @@ class Job {
    * extend the "process" method instead.
    */
   async start (client = db) {
+    Log.debug('JOB ' + this.constructor.name + ' START')
```
sgiaccio: Why not use template string?

server/job/job.js
```diff
@@ -81,6 +85,8 @@ class Job {
         throw new Error('Job canceled or errors found; rollback transaction')
       }
     })
+
+    Log.debug('JOB ' + this.constructor.name + ' END')
```
sgiaccio: Same as in line 60 (template string)

PR #167, opened 2019-05-02
Record nodes access optimization

fefrain:


common/record/record.js
```diff
@@ -212,16 +229,84 @@ const deleteNode = node =>
...
 
     return R.reduce(
       (recordCurrent, child) => deleteNode(child)(recordCurrent),
-      recordValidationUpdated,
+      recordCacheUpdated,
       children
     )
   }
 
+// ======= CACHE
```
minotogna: please, document the data structure and methods

common/record/record.js
```diff
@@ -212,16 +229,84 @@ const deleteNode = node =>
...
+
```
minotogna: Record.js is becoming huge, would it make sense to move the cache part into a separate file, e.g. RecordCache or .. (that is accessible only from within record.js)?

server/modules/record/persistence/recordManager.js
```diff
@@ -1,11 +1,12 @@
 const db = require('../../../db/db')
 
+const Record = require('../../../../common/record/record')
+const { toUuidIndexedObj } = require('../../../../common/survey/surveyUtils')
```
minotogna: should  we use  ```const SurveyUtils = require('../../../../common/survey/surveyUtils')``` ?

common/record/record.js
fefrain: yes we can

common/record/recordCache.js
```diff
@@ -0,0 +1,83 @@
+const R = require('ramda')
```
minotogna: Please, comment data structure and methods.

common/record/recordCache.js
```diff
@@ -0,0 +1,83 @@
...
+
```
minotogna: should we move the file into _internal folder?

PR #170, opened 2019-05-03
Sort sql injection fix

sgiaccio:


common/dataSort.js
```diff
@@ -0,0 +1,90 @@
+const R = require('ramda')
+
+const keys = {
+  order: {
+    asc: 'asc',
+    desc: 'desc',
+  },
+}
+
+const serialize = sortCriteria =>
```
minotogna: is this method still used?

common/dataSort.js
```diff
@@ -0,0 +1,90 @@
...
+  R.pipe(
+    R.map(c => `${c.variable} ${c.order}`),
+    R.join(',')
+  )(sortCriteria)
+
+const getHttpParam = sortCriteria =>
```
minotogna: since you are using this method twice with ```JSON.stringify(DataSort.getHttpParam(sort))``` would it make sense to call this method ```DataSort.toHttpParams``` and returns already the stringified version of it?

common/dataSort.js
```diff
@@ -0,0 +1,90 @@
+const R = require('ramda')
+
```
minotogna: I don't think it's a good idea to mix dataSort object (which is a UI model object) in common folder. can it be moved back where it was?
You can then move the ```getSortPreparedStatement``` into the TableViewQuery or create a module that does the conversion into the same folder of TableViewQuery

minotogna: ok  let's put this in common/surveyRdb

common/dataSort.js
sgiaccio: Not used anymore, deleting it

common/dataSort.js
sgiaccio: Ok, done

common/dataSort.js
sgiaccio: Ok

common/surveyRdb/dataSort.js
```diff
@@ -0,0 +1,82 @@
...
+
+const getViewExpr = R.pipe(
+  R.map(c => `${c.label} ${c.order === keys.order.asc ? 'ASCENDING' : 'DESCENDING'}`),
+  R.join(', ')
+)
+
+const getNewCriteria = availableVariables => sortCriteria =>
+  R.filter(c => R.any(v => v.value === c.variable, availableVariables))(sortCriteria)
+
+const updateOrder = (pos, order) => sortCriteria => {
```
minotogna: can this become directly:
```const updateOrder = (pos, order) => R.assocPath([pos,'order'], order)``` ?

File: common/exprParser/expression.js [hunk 5]
```diff
@@ -3,10 +3,9 @@ const R = require('ramda')
 const jsep = require('./helpers/jsep')
 const { evalExpression } = require('./helpers/evaluator')
 const { toString: toStringUtils, isValid } = require('./helpers/utils')
-const { types } = require('./helpers/types')
+const { types } = require('./types')
```
minotogna: there's no need to move types. it was intentionally left inside the private 'helpers' folder.
It is exported from Expression. So to access it from outside, you can use  ```Expression.types```

common/surveyRdb/dataSort.js
```diff
@@ -0,0 +1,82 @@
...
+
+const getVariable = (variables, value) =>
+  R.find(v => v.value === value)(variables)
+
+const getViewExpr = R.pipe(
+  R.map(c => `${c.label} ${c.order === keys.order.asc ? 'ASCENDING' : 'DESCENDING'}`),
+  R.join(', ')
+)
+
+const getNewCriteria = availableVariables => sortCriteria =>
```
minotogna: can this become ```const getNewCriteria = availableVariables =>  R.filter(c => R.any(v => v.value === c.variable, availableVariables))``` ?

common/surveyRdb/dataSort.js
```diff
@@ -0,0 +1,82 @@
...
+    return {
+      clause: `${prev.clause}${i ? ', ' : ''} $/${paramName}:name/ ${order}`,
+      params: { ...prev.params, [paramName]: curr.variable },
+    }
+  }, { clause: '', params: {} })
+}
+
+const getVariable = (variables, value) =>
+  R.find(v => v.value === value)(variables)
+
```
minotogna: can this become ```const findVariableByValue =  value =>R.find(R.propEq('value', value))```?

common/surveyRdb/dataSort.js
```diff
@@ -0,0 +1,82 @@
...
+const updateOrder = (pos, order) => sortCriteria => {
+  const newVarSortCriteria = R.pipe(
+    R.nth(pos),
+    R.assoc('order', order)
+  )(sortCriteria)
+
+  return R.update(pos, newVarSortCriteria, sortCriteria)
+}
+
+const updateVariable = (pos, variable) => sortCriteria => {
```
minotogna: can this become 
```const updateVariable = (pos, variable) => R.assoc(pos, variable)```?

File: common/surveyRdb/dataSort.js [hunk 9]
```diff
@@ -0,0 +1,82 @@
...
+    R.assoc('label', variable.label)
+  )(sortCriteria)
+
+  return R.update(pos, newVarSortCriteria, sortCriteria)
+}
+
+const getUnchosenVariables = availableVariables => sortCriteria =>
+  R.reject(v => R.any(s => s.variable === v.value)(sortCriteria))(availableVariables)
+
+const addCriteria = (variable, label, order) => sortCriteria =>
```
minotogna: can this become ```const addCriteria = (variable, label, order) => R.append({ variable, label, order })``` ?

common/surveyRdb/dataSort.js
```diff
@@ -0,0 +1,82 @@
...
+  R.append({ variable, label, order }, sortCriteria)
+
+const deleteCriteria = pos => sortCriteria =>
```
minotogna: can this become ```const deleteCriteria = pos => R.remove(pos, 1)``` ?

common/exprParser/expression.js
sgiaccio: ok

common/surveyRdb/dataSort.js
sgiaccio: yes, better

common/surveyRdb/dataSort.js
sgiaccio: yes

common/surveyRdb/dataSort.js
sgiaccio: yes

common/surveyRdb/dataSort.js
sgiaccio: ok, using R.assoc but have to get the properties of the variable object

common/surveyRdb/dataSort.js
sgiaccio: yes

common/surveyRdb/dataSort.js
sgiaccio: yes

PR #174, opened 2019-05-07
Add error badge to invalid rows in data query (edit mode)

fefrain:


common/validation/validator.js
```diff
@@ -208,7 +208,9 @@ const mergeValidation = validation =>
     getValidation,
     R.mergeDeepLeft(validation),
     cleanup,
-    v => assocValidation(v)(obj)
+    v => {
+      return assocValidation(v)(obj)
+    }
   )(obj)
```
minotogna: can it return to ```v => assocValidation(v)(obj)``` ?

server/modules/surveyRdb/service/surveyRdbService.js
```diff
@@ -65,12 +65,12 @@ const queryTable = async (surveyId, nodeDefUuidTable, tableName, nodeDefUuidCols
...
 
+        const record = await RecordManager.fetchRecordByUuid(surveyId, recordUuid)
+
         const resultRow = {
           ...row,
           cols: {},
-          record: {
-            uuid: recordUuid,
-          },
+          record,
```
minotogna: is it really needed to pass all the record? I think it's better to pass only the validation status. or?

webapp/commonComponents/errorBadge.js
```diff
@@ -34,7 +38,12 @@ const ErrorBadge = ({ validation, showLabel, label }) => {
...
   validation: null,
   showLabel: true,
-  label: 'INVALID'
+  label: 'INVALID',
+  /**
+   * error message to show if validation is invalid
+   * (if not specified, field validation errors will be shown)
+   */
+  tooltipErrorMessage: null
 }
```
minotogna: can you just put one line comment in this case? e.g.
``` //error message to show if validation is invalid - if not specified, field validation errors will be shown```

File: common/validation/validator.js (hunk 1 above)
Comment (fefrain) [reply common/validation/validator.js]: sure

File: server/modules/surveyRdb/service/surveyRdbService.js (hunk 2 above)
Comment (fefrain) [reply server/modules/surveyRdb/service/surveyRdbService.js]: right

File: webapp/commonComponents/errorBadge.js (hunk 3 above)
Comment (fefrain) [reply webapp/commonComponents/errorBadge.js]: ok

File: server/modules/record/service/update/thread/recordUpdateThread.js [hunk 4]
```diff
@@ -53,12 +58,19 @@ class RecordUpdateThread extends Thread {
       this.postMessage({ type, content })
   }
 
-  async handleNodesUpdated (updatedNodes) {
+  async handleNodesUpdated (updatedNodes, t) {
     this._postMessage(WebSocketEvents.nodesUpdate, updatedNodes)
```
minotogna: why are you passing the t parameter here?

server/modules/record/service/update/thread/recordUpdateThread.js
```diff
@@ -53,12 +58,19 @@ class RecordUpdateThread extends Thread {
...
   }
 
-  async handleNodesValidationUpdated (validations) {
-    this._postMessage(WebSocketEvents.nodeValidationsUpdate, validations)
+  async handleNodesValidationUpdated (validations, t) {
+    const record = await this.getRecord(t)
```
minotogna: i don't think it's a good idea having the transaction in the handleCallbacks.. don't you have already the record in the thread?

server/modules/record/service/update/thread/recordUpdateThread.js
fefrain: it's not required, right

PR #184, opened 2019-05-10
Localized sidebar and appView

sgiaccio:


common/i18n/i18nFactory.js
```diff
@@ -0,0 +1,33 @@
+const i18next = require('i18next')
+
+const enTranslation = require('./resources/en')
+
+const createParams = lang => ({
+  fallbackLng: 'en',
+  debug: true,
```
minotogna: Why debug true?

webapp/loggedin/appSideBar/appSideBarFooter.js
```diff
@@ -1,31 +1,42 @@
...
-          ? 'Open Foris' : 'OF'
-      }
-    </a>
-  </div>
-)
+class AppSideBarFooter extends React.PureComponent {
+  render () {
+    const { logout, opened } = this.props
+    const { t } = this.context.i18n
+
```
minotogna: It's not clear when reading just t in the code.
What if we import the ```const i18n = this.context.i18n``` and then use it as```i18n.t('key')```?

webapp/loggedin/appView.js
```diff
@@ -1,28 +1,46 @@
...
+
+      </div>
+    )
+  }
+}
 
-    </div>
-  )
+const mapStateToProps = (state) => {
+  return { i18n: AppState.getState(state).i18n }
```
minotogna: Let's create a method in AppState.getI18n?

webapp/loggedin/modules/home/homeView.js
```diff
@@ -43,6 +45,7 @@ class HomeView extends React.Component {
 
   render () {
     const { surveyInfo } = this.props
+    const { t } = this.context.i18n
```
minotogna: As above

common/i18n/i18nFactory.js
sgiaccio: Just for the time being, for development... but right should be removed on master

webapp/loggedin/appSideBar/appSideBarFooter.js
sgiaccio: ok

webapp/loggedin/modules/home/homeView.js
sgiaccio: ok

webapp/loggedin/appView.js
sgiaccio: ok

common/i18n/i18nFactory.js
minotogna: you could use https://github.com/openforis/arena/blob/481d40c0da07e4dc1327761f859cb6398562c318/common/processUtils.js#L5

webapp/survey/surveyInfo/reducer.js
```diff
@@ -2,7 +2,7 @@ import { exportReducer } from '../../utils/reduxUtils'
 
 import Survey from '../../../common/survey/survey'
 
-import { assocSurveyInfoProp, assocSurveyInfoValidation, markDraft, markPublished } from './surveyInfoState'
+import { assocSurveyInfoProp, assocSurveyInfoValidation, markDraft } from './surveyInfoState'
```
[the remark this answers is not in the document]
sgiaccio: ok

common/i18n/i18nFactory.js
sgiaccio: ok, done

PR #185, opened 2019-05-10
Record nodes access optimization

fefrain:


common/record/_internal/recordNodesIndex.js
```diff
@@ -74,60 +91,84 @@ const addNodes = nodes =>
...
       R.values,
       R.reduce(
-        (acc, node) => Node.isDeleted(node)
-          ? acc
-          : _addNode(node)(acc),
+        (recordAcc, node) =>
+          R.ifElse(
+            R.always(Node.isDeleted(node)),
+            R.identity,
+            _addNode(node)
```
minotogna: Can this just be ```Node.isDeleted(node)? _addNode(node) : recordAcc```

common/record/_internal/recordNodesIndex.js
```diff
@@ -74,60 +91,84 @@ const addNodes = nodes =>
...
+    R.always(add),
+    R.ifElse(
+      R.includes(value),
+      R.identity,
+      R.append(value)
+    ),
+    R.without(value)
+  ),
+  arr => R.assocPath(path, arr, obj)
+)(obj)
```
minotogna: It's not very clear to me have a method that does both. Can we have two? Like ```_assocToIndexPath``` and ```_dissocToIndexPath```?

common/record/_internal/recordReader.js
```diff
@@ -141,55 +125,32 @@ const getDependentNodes = (survey, node, dependencyType) => record => {
...
-      }
-    )
+    for (const ancestor of ancestors) {
+      const children = getNodeChildren(ancestor)(record)
+      const nodeFound = children.find(node => Node.getNodeDefUuid(node) === parentCodeDefUuid)
+      if (nodeFound)
+        return nodeFound
+    }
   }
-
```
minotogna: Why has this changed?

common/record/_internal/recordUpdater.js
```diff
@@ -13,17 +13,20 @@ const assocNodes = nodes =>
...
-    const dirtyNodes = RecordReader.findNodesIndexed(Node.isDirty)(record)
+    const dirtyNodeUuids = NodesIndex.getNodeUuidsDirty(record)
 
     const nodesToUpdate = R.pipe(
       R.filter(
         n => {
-          const dirtyNode = R.prop(Node.getUuid(n), dirtyNodes)
+          const nodeUuid = Node.getUuid(n)
+          const dirtyNode = R.includes(nodeUuid, dirtyNodeUuids) && RecordReader.getNodeByUuid(nodeUuid)(record)
+
```
minotogna: If is in the dirtyNodes index, it is also in the record or isn't it?

common/record/node.js
```diff
@@ -27,12 +27,12 @@ const metaKeys = {
   hierarchy: 'h',
   childApplicability: 'childApplicability',
   defaultValue: 'defaultValue',
+  codeAttributeHierarchy: 'codeAttrH',
 }
```
minotogna: Should we call these keys hierarchy and hierachyCode?

common/record/_internal/recordNodesIndex.js
fefrain: right, it would be more clear

common/record/_internal/recordNodesIndex.js
fefrain: ok

PR #197, opened 2019-05-16
I18n advanced props

sgiaccio:


webapp/loggedin/surveyViews/nodeDefEdit/basic/basicProps.js
```diff
@@ -52,13 +28,41 @@ const BasicProps = props => {
...
+      key: NodeDefLayout.nodeDefDisplayIn.ownPage,
+      label: i18n.t('nodeDefEdit.basicProps.ownPage'),
+    },
+  ]
+
+  const onPropLabelsChange = (labelItem, key, currentValue) => {
+    putNodeDefProp(nodeDef, key, R.assoc(labelItem.lang, labelItem.label, currentValue))
+  }
+
   return (
```
minotogna: since displayAsItems and displayInItems are in a way not part of the component, i would have left them outside and converted them as a method . e.g. ```getDisplayInItems = i18n => ([..., ...])``` what do you think?

PR #202, opened 2019-05-17
CollectImportJob - do not publish survey

fefrain:


server/modules/taxonomy/repository/taxonomyRepository.js
```diff
@@ -240,26 +240,27 @@ const deleteDraftTaxaByTaxonomyUuid = async (surveyId, taxonomyUuid, client = db
...
- */
-const getPropCol = (propName, draft, columnPrefix = '') =>
-  `${columnPrefix} ${draft ? 'props_draft' : 'props'}->>'${propName}'`
-
 /**
  * Combines a draft and a published column prop, if needed, using the COALESCE function
  */
-const getPropDraftOrNot = (propName, draft, columnPrefix) =>
-  draft
+const getPropColCombined = (propName, draft, columnPrefix) => {
```
minotogna: nice. should this go to dbUtils? or not?

server/modules/taxonomy/repository/taxonomyRepository.js
```diff
@@ -240,26 +240,27 @@ const deleteDraftTaxaByTaxonomyUuid = async (surveyId, taxonomyUuid, client = db
...
+  /**
+   * If draft, returns something like prop_draft->>'propName', otherwise prop->>'propName'
+   */
+  const getPropCol = (propName, draft, columnPrefix = '') =>
+    `${columnPrefix}${draft ? 'props_draft' : 'props'}->>'${propName}'`
+
+  return draft
     ? `COALESCE(${getPropCol(propName, true, columnPrefix)}, ${getPropCol(propName, false, columnPrefix)})`
     : getPropCol(propName, false, columnPrefix)
+}
```
minotogna: should this be props_draft || props to merge the props when fetching draft?

fefrain: No, you don't have to merge them always. Suppose that you have a published survey with some changes in draft and you are loading only the published version, you don't have to include the draft changes into it.

minotogna: correct, I meant only when draft = true. 
 so when draft = true , ```col = props || props_draft ```, when draft = false ```col = props```.
Otherwise,  when you pass draft=true,  don't you risk to exclude the prop you want to fetch when it is in props and not in props_draft?

server/modules/taxonomy/repository/taxonomyRepository.js
fefrain: right, I'll move it

server/modules/taxonomy/repository/taxonomyRepository.js
fefrain: good, many thanks, I'll simplify that function

PR #212, opened 2019-05-21
Removed surveyIndex, added surveyRefDataIndex

minotogna:


server/modules/collectImport/service/collectImport/collectImportJob.js
```diff
@@ -8,7 +8,6 @@ const CategoriesImportJob = require('./metaImportJobs/categoriesImportJob')
 const TaxonomiesImportJob = require('./metaImportJobs/taxonomiesImportJob')
 const NodeDefsImportJob = require('./metaImportJobs/nodeDefsImportJob')
 const RecordsImportJob = require('./dataImportJobs/recordsImportJob')
-const SurveyIndexGeneratorJob = require('../../../survey/service/surveyIndexGeneratorJob')
 const SurveyRdbGeneratorJob = require('../../../surveyRdb/service/surveyRdbGeneratorJob')
 const RecordCheckJob = require('../../../survey/service/recordCheckJob')
```
sgiaccio: RecordCheckJob import is unused, should it be commented out until RecordCheckJob is restored?

common/survey/_internal/surveyRefDataIndex.js
```diff
@@ -0,0 +1,160 @@
...
+  categoryItemUuidIndex: 'categoryItemUuidIndex',
+  categoryItemIndex: 'categoryItemIndex',
+  taxonUuidIndex: 'taxonUuidIndex',
+  taxonIndex: 'taxonIndex'
+}
+
+// ====== READ
+
+// ==== category index
+const getCategoryItemUuidAndCodeHierarchy = (survey, nodeDef, record, parentNode, code) => surveyIndex => {
```
sgiaccio: Why not split this into two separate functions? (getCategoryItemUuid and getCodeHierarchy)?

server/modules/survey/manager/surveyManager.js
```diff
@@ -93,12 +92,22 @@ const fetchSurveyById = async (surveyId, draft = false, validate = false, client
...
+const fetchSurveyAndNodeDefsAndRefDataBySurveyId = async (surveyId, draft = false, advanced = false, validate = false, client = db) => {
+  const survey = await fetchSurveyAndNodeDefsBySurveyId(surveyId, draft, advanced, validate, client)
+  const [categoryIndexRS, taxonomyIndexRS] = await Promise.all([
+    CategoryRepository.fetchIndex(surveyId, draft, client),
+    TaxonomyRepository.fetchIndex(surveyId, draft, client)
+  ])
+
+  return Survey.assocRefData(categoryIndexRS, taxonomyIndexRS)(survey)
+}
+
```
sgiaccio: Why not split the two functions above into two and three separate functions?

server/modules/collectImport/service/collectImport/collectImportJob.js
minotogna: yes it's temporary left there. [there's another task to get rid of it in collectImportJob](https://trello.com/c/Fk8LbB0F/185-refactor-collectimportjob-surveypublishjob-recordcheckjob) .

common/survey/_internal/surveyRefDataIndex.js
minotogna: to avoid duplicating the call to ```RecordReader.getParentCodeAttriubute``` which could be time consuming. since the results are used only together from outside, we thought to do once only the heavy computational logic

server/modules/survey/manager/surveyManager.js
minotogna: Because the SurveyManger responsibility is to compose the objects to be consumed by outside callers.

PR #216, opened 2019-05-21
I18n categories

sgiaccio:


common/i18n/resources/en.js
```diff
@@ -184,4 +192,38 @@ module.exports = {
...
+    level: 'Level',
+    items: 'Items',
+    confirmDelete: 'Delete the item with all children? $t(common.cantUndoWarning)',
+    deleteItem: 'Delete item',
+  },
+
+  jobs: {
+    item: 'Item',
+    errors: 'Errors',
+
```
minotogna: should item and error plus plural go to common? what do you think?

webapp/loggedin/surveyViews/items/itemsTable.js
```diff
@@ -21,7 +24,7 @@ const TableRow = props => {
       <div className="name">
         {name}
         <ErrorBadge validation={item.validation}/>
-        <WarningBadge show={!item.usedByNodeDefs} label="Unused"/>
+        <WarningBadge show={!item.usedByNodeDefs} label={i18n.t('unused')}/>
```
minotogna: is the key maybe ```itemsTable.unused``` ?

sgiaccio: Yes, fixed

common/i18n/resources/en.js
sgiaccio: yes, better

PR #217, opened 2019-05-22
Record validation refactor

minotogna:


package.json
```diff
@@ -2,7 +2,7 @@
   "license": "MIT",
   "name": "open-foris-arena",
   "engines": {
-    "node": ">=10.13.0",
+    "node": "10.15.3",
```
fefrain: isn't better to use >=10.5.3 instead?

server/modules/record/manager/_recordManager/recordUniquenessValidator.js
```diff
@@ -49,19 +47,19 @@ const validateEntityKeysUniqueness = (survey, record, nodeEntity) => {
   )(entityValidations)
 }
 
-const isDuplicateEntity = (survey, record, siblingEntitiesAndSelf, entity) => {
+const _isEntityDuplicated = (survey, record, siblingEntitiesAndSelf, entity) => {
```
fefrain: it should be "_isEntityDuplicate" not "duplicateD"...

minotogna: not sure: http://conjugator.reverso.net/conjugation-english-verb-duplicate.html

package.json
minotogna: I'm not sure; but i have changed it. see next commit

server/modules/record/manager/_recordManager/recordUniquenessValidator.js
fefrain: duplicate is an [adjective](https://www.dictionary.com/browse/duplicated)!

minotogna: changed!

PR #233, opened 2019-06-04
validate records uniqueness in EntitiesUniquenessValidationJob

fefrain:


common/i18n/resources/en.js
```diff
@@ -228,10 +228,10 @@ module.exports = {
     CategoriesValidationJob: 'Categories Validation',
     CollectImportJob: 'Collect Import',
     CollectSurveyReaderJob: 'Collect Survey Reader',
-    EntitiesUniquenessValidationJob: 'Entities Uniqueness Validation',
     NodeDefsImportJob: 'Node Defs Import',
     NodeDefsValidationJob: 'Node Defs Validation',
     RecordCheckJob: 'Record Check',
+    RecordsAndEntitiesUniquenessValidationJob: 'Records and Entities Uniqueness Validation',
```
minotogna: Why having this long name RecordsAndEntitiesUniquenessValidationJob. Could it be like before EntitiesUniquenessValidationJob, and we know that if a root entity is duplicate, it means there's another record with same error?

fefrain: good point, I'll change it.

PR #234, opened 2019-06-05
I18n errors

sgiaccio:


server/modules/collectImport/service/collectImport/metaImportJobs/taxonomiesImportJob.js
```diff
@@ -149,8 +148,8 @@ class TaxonomiesImportJob extends Job {
       this.addError({
         [keysError.code]: {
           valid: false,
-          errors: [`duplicate code ${code}; row: ${this.currentRow} duplicate row: ${rowDuplicateCode}`]
-        }
+          errors: [{ key: 'duplicateCode', params: { code, row: this.currentRow, duplicateRow: rowDuplicateCode } }],
```
minotogna: it would be preferable to have duplicateCode to keysError constant as the other key errors

server/modules/collectImport/service/collectImport/metaImportJobs/taxonomiesImportJob.js
```diff
@@ -162,8 +161,8 @@ class TaxonomiesImportJob extends Job {
       this.addError({
         [keysError.scientificName]: {
           valid: false,
-          errors: [`duplicate scientific name ${scientificName}; row: ${this.currentRow} duplicate row: ${rowDuplicateScientificName}`]
-        }
+          errors: [{ key: 'duplicateName', params: { scientificName, row: this.currentRow, duplicateRow: rowDuplicateScientificName } }],
```
minotogna: it would be preferable to have duplicateName to keysError constant as the other key errors

common/i18n/resources/en.js
```diff
@@ -245,4 +245,26 @@ module.exports = {
     TaxonomiesValidationJob: 'Taxonomies Validation',
     TaxonomyImportJob: 'Taxonomy Import',
   },
+
+  formErrors: {
+    duplicateName: 'duplicate scientific name {{scientificName}}; row: {{row}} duplicate row: {{duplicateRow}}',
+    duplicateCode: 'duplicate code {{code}}; row: {{row}} duplicate row: {{duplicateRow}}',
```
fefrain: can we use a nested key in this case? something like $t(duplicateRow)...

webapp/commonComponents/validationTooltip.js
```diff
@@ -0,0 +1,24 @@
...
+import useI18n from '../commonComponents/useI18n'
+
+export default ({ validation, children, className, type = '' }) => {
+  const i18n = useI18n()
+
+  const messagesHtml = ValidationUtils.getValidationFieldMessagesHTML(i18n)(validation)
+
+  return (
+    <Tooltip
+      type={type}
```
minotogna: Should ValidationTooltip always use ```type``` error?

sgiaccio: In theory yes, but the inner element (input and checkbox) get a different style when the tooltip class is tooltip-error - see tooltip.scss line 30. We could set an "error" class in the className instead?

server/modules/collectImport/service/collectImport/metaImportJobs/taxonomiesImportJob.js
sgiaccio: Yes right

server/modules/collectImport/service/collectImport/metaImportJobs/taxonomiesImportJob.js
sgiaccio: Yes

webapp/commonComponents/validationTooltip.js
minotogna: maybe ValiadationTooltip should conditionally pass the type param (error) to tooltip if validation is not valid?

sgiaccio: ok done

common/i18n/resources/en.js
sgiaccio: ok done

webapp/commonComponents/validationTooltip.js
```diff
@@ -0,0 +1,25 @@
...
+
+import Tooltip from './tooltip'
+
+import useI18n from '../commonComponents/useI18n'
+
+export default ({ validation, children, className }) => {
+  const i18n = useI18n()
+
+  const messagesHtml = ValidationUtils.getValidationFieldMessagesHTML(i18n)(validation)
+  const type = R.isEmpty(validation) ? '' : 'error'
```
minotogna: should this condition include ```Validator.isValidationValid```  as well?
i.e. ```R.isEmpty(validation) || Validator.isValidationValid(validation) ? '' : 'error' ```

fefrain: I think that the condition R.isEmpty(validation) || Validator.isValidationValid(validation) should be evaluated before generating the messageHtml, because in that case there won't be any messages to show.

sgiaccio: removed R.isEmpty(validation) and passing validation from errorBadge; Checking validity before generating messagesHtml

PR #237, opened 2019-06-05
show all node def hierarchy in report item header

fefrain:


webapp/loggedin/modules/home/collectImportReport/nodeDefReportItem.js
```diff
@@ -69,4 +70,14 @@ const NodeDefReportItem = props => {
   )
 }
 
+const _getNodeDefPath = (survey, nodeDef, lang) => {
+  const nodeDefPathParts = []
+  Survey.visitAncestorsAndSelf(
+    nodeDef,
+    def => nodeDefPathParts.unshift(NodeDef.getLabel(def, lang) || NodeDef.getName(def))
```
minotogna: ```NodeDef.getLabel``` already returns the name when the label hasn't been specified

fefrain: it's true...

PR #241, opened 2019-06-07
Deleted getDefaultLanguage from client

sgiaccio:


webapp/commonComponents/expression/expressionEditorPopup.js
```diff
@@ -15,6 +15,8 @@ import * as SurveyState from '../../survey/surveyState'
 import * as ExpressionVariables from './expressionVariables'
 import * as ExpressionParser from './expressionParser'
 
+import appContext from '../../app/appContext'
```
minotogna: please let's use uppercase convention as usual

common/i18n/resources/en.js
```diff
@@ -246,10 +247,10 @@ module.exports = {
     TaxonomyImportJob: 'Taxonomy Import',
   },
 
-  formErrors: {
+  errors: {
```
minotogna: let's keep separate formErrors to systemErrors (as agreed)

webapp/commonComponents/expression/expressionEditorPopup.js
sgiaccio: Done

common/i18n/resources/en.js
sgiaccio: Ok

PR #249, opened 2019-06-26
Ui tabbar remove

minotogna:


webapp/loggedin/appSideBar/appSideBarModules.js
```diff
@@ -81,9 +81,9 @@ const AppSideBarModules = ({ pathname, surveyInfo, sideBarOpened }) => (
   <div className="app-sidebar__modules">
     {
       modulesHierarchy.map((module, i) => (
-        <React.Fragment>
+        <React.Fragment key={i}>
```
sgiaccio: Why not key={module.key}?

minotogna: because it's the same.. i is shorter... but i updated it in the next commit

PR #259, opened 2019-07-09
Survey create refactor

minotogna:


webapp/loggedin/modules/home/surveyList/surveyListView.js
```diff
@@ -1,42 +1,54 @@
...
+    setFeched(true)
+  }, [])
+
+  const surveysLength = R.length(surveys)
+  useEffect(() => {
+    // redirect to create survey when (after fetching surveys) there are no surveys
+    if (fetched === true && surveysLength === 0) {
+      history.push(appModuleUri(homeModules.surveyNew))
+    }
+  }, [surveysLength])
```
sgiaccio: Should it be [surveysLength, fetched]?

minotogna: It does't matter. at that point is always true, because the surveys have been fetched, therefore the surveysLength has changed.

PR #261, opened 2019-07-10
i18n-ed expression editor

sgiaccio:


webapp/commonComponents/expression/expressionEditorPopup.js
```diff
@@ -58,7 +58,8 @@ class ExpressionEditorPopup extends React.Component {
       isBoolean, onChange, onClose,
     } = this.props
```
minotogna: Reset Line 93, Apply line 99 are missing

webapp/commonComponents/expression/nodes/logical.js
```diff
@@ -3,13 +3,17 @@ import * as R from 'ramda'
 
 import Expression from '../../../../common/exprParser/expression'
```
minotogna: group - line 39 is missing

PR #262, opened 2019-07-10
I18n fixes

sgiaccio:


common/i18n/resources/en.js
```diff
@@ -1,39 +1,43 @@
...
     error: 'error',
     error_plural: 'Errors',
-    item: 'Item',
-    item_plural: 'Items',
     errorMessage: 'Error message',
     errorMessage_plural: 'Error messages',
     invalid: 'INVALID',
+    item: 'Item',
+    item_plural: 'Items',
+    label: 'Label',
```
minotogna: can we put label_plurarl here?

sgiaccio: Done

PR #288, opened 2019-08-05
Record validator bug fixes

fefrain:


webapp/loggedin/surveyViews/surveyForm/nodeDefs/components/nodeDefTableCellBody.js
```diff
@@ -29,15 +29,11 @@ const getNodeValues = (surveyInfo, nodeDef, nodes, lang) => {
...
-    '',
-    nodes
-  )
-
+  return R.pipe(
+    R.reject(node => Node.isPlaceholder(node) || Node.isValueBlank(node)),
+    R.map(getNodeValue),
+    R.join(', ')
+  )(nodes)
 }
```
minotogna: Reduce iterates only once, with the pipe you added three times. Is it needed?

webapp/loggedin/surveyViews/surveyForm/nodeDefs/components/nodeDefErrorBadge.js
```diff
@@ -59,8 +59,13 @@ const mapStateToProps = (state, props) => {
       // showing validation for a single node instance of multiple nodeDef
       if (node) {
         validation = RecordValidation.getNodeValidation(node)(recordValidation)
-      } else {
+      } else if (NodeDef.isEntity(nodeDef)) {
```
minotogna: Can we add a comment here?

common/i18n/resources/en.js
```diff
@@ -256,6 +256,7 @@ module.exports = {
   },
 
   formErrors: {
+    atLeastOneInvalidValue: 'At least one invalid value',
```
minotogna: Can the message be something like : one or more values are invalid. What do you think?

PR #294, opened 2019-08-07
Survey validator

minotogna:


common/survey/_survey/surveyDefaults.js
```diff
@@ -1,7 +1,7 @@
 const { groupNames, permissions } = require('../../auth/authGroups')
 const { keys: authGroupKeys } = require('../../auth/authGroups')
 
-const getDefaultAuthGroups = lang => [
+const defaultAuthGroups = [
```
fefrain: can it be a function that generates a new object every time? (like you did for Validator.newValidationValid?)

minotogna: it can, but it's used only when inserting a survey. basically only to insert those groups into the db. do you think the function is needed?

minotogna: I have restored the getDefaultAuthGroups method

PR #296, opened 2019-08-08
Survey edit info bug fix

fefrain:


server/modules/nodeDef/manager/nodeDefManager.js
```diff
@@ -53,6 +55,49 @@ const updateNodeDefProps = async (user, surveyId, nodeDefUuid, props, propsAdvan
     return nodeDef
   })
 
+const deleteNodeDefValidationMessagesLabels = async (surveyId, deletedLanguages, client = db) => {
+  await client.tx(async tx => {
```
minotogna: what about calling this method ```deleteValidationMessageLabels```? ( I know, we'll eventually remove the nodeDef prefix to other methods as well)

server/modules/nodeDef/manager/nodeDefManager.js
```diff
@@ -53,6 +55,49 @@ const updateNodeDefProps = async (user, surveyId, nodeDefUuid, props, propsAdvan
...
+          // 4. remove messages in deleted languages
+          const messagesUpdated = R.omit(deletedMessageLanguages, messages)
+          expressionsUpdated.push(NodeDefExpression.assocMessages(messagesUpdated)(expression))
+          nodeDefUpdateRequired = true
+        }
+
+        if (nodeDefUpdateRequired) {
+          // 5. update node def in db (props already published)
+          const validationsUpdated = NodeDefValidations.assocExpressions(expressionsUpdated)(validations)
+          await NodeDefRepository.updateNodeDefPropsPublished(surveyId, NodeDef.getUuid(nodeDef), {}, {
```
minotogna: why did you introduce this ```NodeDefRepository .updateNodeDefPropsPublished```? Isn't the publishProps happening after this? don't you just need to update the expressions draft props here?

server/modules/nodeDef/manager/nodeDefManager.js
fefrain: Now it's consistent with the other names, we will change it in the future.

server/modules/nodeDef/manager/nodeDefManager.js
fefrain: in the other cases (survey labels/descriptions, nodeDef labels/descriptions, category items labels) it's easier to do it after survey props are published, so it's better to remove even the validation message labels together with them.
Also, NodeDefRepository.updateNodeDefProps updates even the date_modified, and it's not needed in this case.

server/modules/nodeDef/manager/nodeDefManager.js
minotogna: I'm confident this huge method can be replaced by a few lines of an SQL queries, within publishProps method. Let's try tomorrow

PR #297, opened 2019-08-08
App notification

minotogna:


webapp/loggedin/modules/users/userInvite/userInviteView.js
```diff
@@ -16,7 +16,7 @@ import Dropdown from '../../../../commonComponents/form/dropdown'
 import * as AppState from '../../../../app/appState'
 import * as SurveyState from '../../../../survey/surveyState'
 
-import { showAppLoader, hideAppLoader } from '../../../../app/actions'
+import { showAppLoader, hideAppLoader, showNotificationMessageKey } from '../../../../app/actions'
```
sgiaccio: Why is showNotificationMessageKey imported?

minotogna: Because after having invited the user, the application shows a notification message. See userInviteState

PR #298, opened 2019-08-08
User UUID

sgiaccio:


test/testContext.js
```diff
@@ -15,7 +15,7 @@ let survey = null
  * before executing all tests
  */
 const initTestContext = async () => {
-  user = await UserService.fetchUserByCognitoUsername('demo_user')
+  user = await UserService.fetchUserByUuid('demo_user')
 }
```
minotogna: Is the uuid demo_user?

fefrain: I don't think so...! That was the username...

minotogna: so are the tests working??

sgiaccio: Fixed, they work now

PR #305, opened 2019-08-13
SurveyInfo validation

fefrain:


server/modules/survey/api/surveyApi.js
```diff
@@ -49,7 +49,7 @@ module.exports.init = app => {
       const surveyId = Request.getRestParam(req, 'surveyId')
       const draft = Request.getBoolParam(req, 'draft')
 
-      const survey = await SurveyService.fetchSurveyById(surveyId, draft)
+      const survey = await SurveyService.fetchSurveyById(surveyId, draft, true)
```
minotogna: why are you always validating? Shouldn't it be validated only if current user can edit survey?

fefrain: no, you are right, as usual

PR #315, opened 2019-08-16
Edit show user

sgiaccio:


webapp/loggedin/modules/users/user/userView.scss
```diff
@@ -0,0 +1,36 @@
+.user-invite {
+    display: grid;
```
minotogna: Is the class user_invite still used?

sgiaccio: No, deleted

PR #318, opened 2019-08-21
Record uniqueness validator

fefrain:


server/modules/record/manager/_recordManager/recordValidationManager.js
```diff
@@ -59,9 +64,22 @@ const persistValidation = async (survey, record, nodesValidation, tx) => {
   await RecordRepository.updateValidation(surveyId, Record.getUuid(record), recordValidationUpdated, tx)
 }
 
+const validateOtherRecordsUniquenessAndPersistValidation = async (survey, record, excludeRecordFromCount, t) => {
```
minotogna: should we call this method validateRecordsUniquenessAndPersistValidation?

server/modules/record/manager/_recordManager/recordUpdateManager.js
```diff
@@ -93,28 +93,57 @@ const deleteRecordsPreview = async (surveyId) =>
...
+      const { record: updatedRecord, updatedNodesAndDependents } = await _onNodesUpdate(
+        survey,
+        nodesUpdated,
+        nodesUpdateListener,
+        nodesValidationListener,
+        t,
+      )
+      await _afterNodesUpdate(user, survey, updatedRecord, updatedNodesAndDependents, t)
+
+      return updatedRecord
```
minotogna: persistNode and deleteNode do the same thing, expect for the persist/delete node. should we have just a method with all the before and after update logic? e.g ```_updateNodeAndValidateKeys```?

server/modules/record/manager/_recordManager/recordUpdateManager.js
```diff
@@ -93,28 +93,57 @@ const deleteRecordsPreview = async (surveyId) =>
  */
```
minotogna: should the sequence diagram be updated as well?

server/modules/record/manager/_recordManager/recordValidationManager.js
fefrain: ok, the shorter the better...

server/modules/record/manager/_recordManager/recordUpdateManager.js
fefrain: or something similiar, right

server/modules/record/manager/_recordManager/recordUpdateManager.js
fefrain: of course

server/modules/record/manager/_recordManager/recordUpdateManager.js
minotogna: why not to call the method for what it really does? the word perform doesn't really tell me what the method does.. don't you think so?

PR #319, opened 2019-08-21
Password reset

sgiaccio:


webapp/app/cognitoAuth.js
```diff
@@ -74,14 +103,9 @@ export const getJwtToken = () => new Promise((resolve, reject) => {
         reject(err)
       }
 
-      try {
-        const accessToken = session.getAccessToken()
-        const jwtToken = accessToken.jwtToken
-        resolve(jwtToken)
-      } catch (e) {
-        reject(e)
```
minotogna: any reason you removed the try catch here?

webapp/app/cognitoAuth.js
```diff
@@ -61,6 +63,33 @@ export const logout = () => {
...
+  new Promise((resolve, reject) => {
+    // Override Cognito error messages with more readable ones
+    if (!(new RegExp(/^[\S]+$/)).test(verificationCode)) {
+      throw new Error('Please enter a valid verification code')
+    } else if (!password) {
+      throw new Error('Please specify a new password')
+    } else if (!(new RegExp(/^[\S]+.*[\S]+$/)).test(password)) {
+      throw new Error('Password should not start nor end with white spaces')
+    } else if (password.length < 6) {
+      throw new Error('Password does not conform to policy: Password not long enough')
```
minotogna: should this message say: 'Password should be at least 6 characters long' . better to show how many characters are required.

webapp/login/actions.js
```diff
@@ -42,13 +46,35 @@ export const acceptInvitation = (name, password) => _createAction(
...
 )
 
-export const forgotPassword = () => dispatch => {
+export const showForgotPasswordForm = () => dispatch => {
   dispatch(setLoginError(null))
-  dispatch(showNotificationMessage('An email with the verification code has been sent'))
+  dispatch({ type: loginUserActionUpdate, action: LoginState.userActions.forgotPassword })
 }
+
+export const sendResetPasswordRequest = email => _createAction(
```
minotogna: should this action be called ```sendVerificationCode```?

webapp/login/actions.js
```diff
@@ -5,9 +5,12 @@ import * as CognitoAuth from '../app/cognitoAuth'
 import * as LoginState from './loginState'
 import { hideAppLoader, initUser, showAppLoader, showNotificationMessage } from '../app/actions'
 
+export const emailUpdate = 'login/email/update'
```
minotogna: i would follow the convention so far used: calling this type loginEmailUpdate

webapp/app/cognitoAuth.js
```diff
@@ -61,6 +63,33 @@ export const logout = () => {
...
+
+export const resetPassword = (verificationCode, password) =>
+  new Promise((resolve, reject) => {
+    // Override Cognito error messages with more readable ones
+    if (!(new RegExp(/^[\S]+$/)).test(verificationCode)) {
+      throw new Error('Please enter a valid verification code')
+    } else if (!password) {
+      throw new Error('Please specify a new password')
+    } else if (!(new RegExp(/^[\S]+.*[\S]+$/)).test(password)) {
+      throw new Error('Password should not start nor end with white spaces')
```
minotogna: should we add a check to verify there's a number and an upper case letter?

webapp/login/actions.js
sgiaccio: right, changed

webapp/app/cognitoAuth.js
sgiaccio: Added unique check that checks password length and content (lowercase, uppercase and numbers)

webapp/app/cognitoAuth.js
sgiaccio: Added unique check that checks password length and content (lowercase, uppercase and numbers)

webapp/app/cognitoAuth.js
sgiaccio: No, it works in both cases but try catch seems more correct...

webapp/login/acceptInvitationForm.js
```diff
@@ -1,43 +1,82 @@
-import React, { useRef } from 'react'
+import React from 'react'
 import { connect } from 'react-redux'
 
+import { Input } from '../commonComponents/form/input'
+import useFormObject from '../commonComponents/hooks/useFormObject'
+import { validate, validateRequired, isValidationValid } from '../../common/validation/validator'
```
minotogna: the objects in common are used without spread import. e.g. ```import Validator from '../../common/validation/validator'```

webapp/login/acceptInvitationForm.js
```diff
@@ -1,43 +1,82 @@
...
+        .find(v => !isValidationValid(v.validation))
+      const key = firstMatch.validation.errors[0].key
+      setLoginError(formErrors[firstMatch.field][key])
     }
   }
 
+  const setUserName = userName => setObjectField('userName', userName)
+  const setPassword = password => setObjectField('password', password)
+  const setPasswordConfirm = passwordConfirm => setObjectField('passwordConfirm', passwordConfirm)
+
```
minotogna: should we move all the component logic in an external file like AcceptInvitationFormState ? it would become more readable.

webapp/login/useResetPasswordFormState.js
```diff
@@ -0,0 +1,67 @@
+import React from 'react'
+
```
minotogna: should we move this file, loginForm, passwordValidator, acceptInvitationForm into an inner folder called components?

PR #326, opened 2019-08-26
Error keys

fefrain:


common/i18n/resources/en.js
```diff
@@ -39,7 +39,7 @@ module.exports = {
     no: 'No',
     noItems: `$t(common.no) $t(common.item,{'count':2})`,
     of: 'of',
-    required: 'Required',
+    valueRequired: 'Required',
```
minotogna: why has the key changed? if the text says 'Required' I would call the key 'required' . note the keys in common are alphabetically ordered . also, is this still used?

common/i18n/resources/en.js
```diff
@@ -285,32 +276,77 @@ module.exports = {
...
+      valueRequired: 'Required value',
+    },
+
+    surveyInfoEdit: {
+      langRequired: 'Language is required',
+      srsRequired: 'Spatial Reference System is required',
+    },
+
+    taxonomyEdit: {
+      codeDuplicate: 'Duplicate code {{code}}; $t(jobErrors.duplicateRows)',
```
minotogna: why do we still have jobErrors?

webapp/utils/validationUtils.js
```diff
@@ -3,42 +3,53 @@ import * as R from 'ramda'
...
+  }
+}
 
-const getErrorText = (error, i18n, errorKeyPrefix = 'formErrors') => error.key === 'custom'
-  ? error.messages[i18n.lang]
-  : i18n.t([`${errorKeyPrefix}.${error.key}`, error.key], error.params)
-
-const getFieldError = (i18n, errorKeyPrefix) => field => R.pipe(
-  R.pathOr([], [field, 'errors']),
+const getFieldError = (i18n, errorKeyPrefix = null) => field => R.pipe(
```
minotogna: do we still need errorKeyPrefix?

common/i18n/resources/en.js
fefrain: right

webapp/utils/validationUtils.js
fefrain: no, I'll remove it

common/i18n/resources/en.js
fefrain: no need for them

PR #330, opened 2019-08-30
Category import

fefrain:


common/survey/categoryImportSummary.js
```diff
@@ -0,0 +1,88 @@
...
+  const headers = getHeaders(summary)
+  return R.pipe(
+    R.keys,
+    R.find(headerName => {
+      const header = headers[headerName]
+      return getHeaderType(header) === type && getHeaderLevelIndex(header) === levelIndex
+    })
+  )(headers)
+}
+
```
minotogna: there's an abuse of the header word. In really ```column``` should be used. header is just a property of the column

common/survey/categoryItem.js
```diff
@@ -12,31 +12,37 @@ const keys = {
...
-      R.prop(language),
-      R.defaultTo(getCode(item))
-    )(item)
+const getLabel = language => item =>
+  R.pipe(
+    getLabels,
+    R.propOr(getCode(item), language),
+  )(item)
+
+const getDescriptions = SurveyUtils.getProp(props.descriptions)
```
minotogna: should we start moving SurveyUtils to ObjectUtils?

common/survey/categoryItem.js
```diff
@@ -12,31 +12,37 @@ const keys = {
...
 const getLabels = SurveyUtils.getProp(props.labels)
 
-const getLabel = language =>
-  item =>
-    R.pipe(
-      getLabels,
-      R.prop(language),
-      R.defaultTo(getCode(item))
-    )(item)
+const getLabel = language => item =>
```
minotogna: don't we have a generic method ```Object.getLabel```? if not, let's add it?

common/survey/categoryItem.js
```diff
@@ -12,31 +12,37 @@ const keys = {
...
+  R.pipe(
+    getLabels,
+    R.propOr(getCode(item), language),
+  )(item)
+
+const getDescriptions = SurveyUtils.getProp(props.descriptions)
+
+const getDescription = language => item =>
```
minotogna: don't we have a generic method ```Object.getDescription```? if not, let's add it?

common/survey/categoryItem.js
```diff
@@ -52,6 +58,12 @@ module.exports = {
   getCode,
   getLabels,
   getLabel,
+  getDescriptions,
+  getDescription,
+  getExtraProp: prop => R.pipe(
+    SurveyUtils.getProp(keys.extra),
```
minotogna: as above, should we start migrating things into ObjectUtils?

server/modules/category/service/categoryImportCSVParser.js
```diff
@@ -0,0 +1,151 @@
...
+  } catch (error) {
+    reject(error)
+  }
+})
+
+const createImportSummary = async (filePath) => {
+  //clean category levels
+  const headers = await _readHeaders(filePath)
+
+  const levelNames = headers.reduce(
```
minotogna: why don't you generate an object with an index as well? e.g. ```[{name:'region',index:0},{name:'district',index:1}]```

server/modules/category/service/categoryImportCSVParser.js
```diff
@@ -0,0 +1,151 @@
...
+      }
+      return accLevelNames
+    },
+    []
+  )
+
+  const headersInfo = headers.reduce(
+    (acc, header) => {
+      if (header.endsWith(columnSuffixes.code)) {
+        const levelIndex = R.pipe(
```
minotogna: if you generate an object (see above comment) you don't need to iterate again on the headers to get it

server/modules/category/service/categoryImportCSVParser.js
```diff
@@ -0,0 +1,151 @@
...
+          R.filter(R.endsWith(columnSuffixes.code)),
+          R.indexOf(header)
+        )(headers)
+
+        const levelName = header.substr(0, header.length - columnSuffixes.code.length)
```
minotogna: don't you have already the level names generated in the first iterations?

server/modules/category/service/categoryImportJob.js
```diff
@@ -0,0 +1,205 @@
...
+    this.logDebug('importing levels')
+
+    const categoryUuid = CategoryImportJobParams.getCategoryUuid(this.params)
+    const summary = CategoryImportJobParams.getSummary(this.params)
+
+    const surveyId = this.getSurveyId()
+    const user = this.getUser()
+
+    await CategoryManager.deleteLevelsByCategory(user, surveyId, categoryUuid, this.tx)
+    let category = await CategoryManager.fetchCategoryByUuid(surveyId, categoryUuid, true, false, this.tx)
```
minotogna: can the ```CategoryManager.deleteLevelsByCategory``` return the category? you won't need the fetch in that case

server/utils/file/csvReader.js
```diff
@@ -0,0 +1,64 @@
...
+      })()
+    }
+
+    const onData = data => {
+      if (headers) {
+        const wasEmpty = queue.isEmpty()
+        queue.enqueue(data)
+        onTotalChange && onTotalChange(++total)
+
+        if (!started || wasEmpty) {
```
minotogna: why also wasEmpty? started is not enough?

common/survey/categoryItem.js
fefrain: This one defaults to the item code... We can't make it generic.

common/survey/categoryItem.js
fefrain: Where else will it be needed?

server/modules/category/service/categoryImportJob.js
fefrain: Right

server/utils/file/csvReader.js
fefrain: There can be the possibility that the queue becomes empty while there are still rows to read.

webapp/loggedin/surveyViews/categoryEdit/actions.js
```diff
@@ -67,6 +69,27 @@ export const createCategory = () => async (dispatch, getState) => {
...
+export const uploadCategory = (categoryUuid, file) => async (dispatch, getState) => {
+  const formData = new FormData()
+  formData.append('file', file)
+
+  const config = {
+    headers: {
+      'content-type': 'multipart/form-data'
+    }
+  }
+
```
sgiaccio: You don't need to set the content-type header, axios does it himself when using FormData

server/modules/category/manager/categoryManager.js
```diff
@@ -43,13 +43,13 @@ const insertCategory = async (user, surveyId, category, client = db) =>
...
-const insertLevel = async (user, surveyId, categoryUuid, level, client = db) =>
+const insertLevel = async (user, surveyId, level, client = db) =>
   await client.tx(async t => {
-    const levelDb = await CategoryRepository.insertLevel(surveyId, categoryUuid, level, t)
+    const levelDb = await CategoryRepository.insertLevel(surveyId, level, t)
 
     await markSurveyDraft(surveyId, t)
 
-    await ActivityLog.log(user, surveyId, ActivityLog.type.categoryLevelInsert, { categoryUuid, level }, t)
+    await ActivityLog.log(user, surveyId, ActivityLog.type.categoryLevelInsert, { level }, t)
```
sgiaccio: Can these three operations be done concurrently? i.e. with a `Promise.all` or

```
const levelDbPromise = CategoryRepository.insertLevel(surveyId, level, t)
const markPromise = markSurveyDraft(surveyId, t)
const logPromise = ActivityLog.log(user, surveyId, ActivityLog.type.categoryLevelInsert, { level }, t)

await markPromise
await logPromise
return await levelDbPromise
```

server/modules/category/manager/categoryManager.js
```diff
@@ -65,6 +65,20 @@ const insertItem = async (user, surveyId, item, client = db) =>
...
+    await CategoryRepository.insertItems(surveyId, items, client)
+
+    await markSurveyDraft(surveyId, t)
+
+    const activities = items.map(item => ({
+        type: ActivityLog.type.categoryItemInsert,
+        params: item
+      })
+    )
+    await ActivityLog.logMany(user, surveyId, activities, client)
```
sgiaccio: Same as above, can these operations be performed concurrently?

server/modules/category/manager/categoryManager.js
```diff
@@ -165,6 +179,14 @@ const deleteLevel = async (user, surveyId, levelUuid) =>
     await ActivityLog.log(user, surveyId, ActivityLog.type.categoryLevelDelete, { levelUuid }, t)
   })
 
+const deleteLevelsByCategory = async (user, surveyId, categoryUuid, client = db) =>
+  await client.tx(async t => {
+    await CategoryRepository.deleteLevelsByCategory(surveyId, categoryUuid, t)
+    await markSurveyDraft(surveyId, t)
+
+    await ActivityLog.log(user, surveyId, ActivityLog.type.categoryLevelsDelete, { categoryUuid }, t)
```
sgiaccio: Same as above, can these operations be performed concurrently?

server/modules/category/manager/categoryManager.js
fefrain: I'll do it with a Promise.all

server/modules/category/manager/categoryManager.js
fefrain: yes, sure

common/survey/categoryItem.js
```diff
@@ -12,31 +12,23 @@ const keys = {
 
 const props = {
   code: 'code',
-  labels: 'labels'
+  descriptions: 'descriptions',
+  extra: 'extra',
+  labels: 'labels',
 }
```
minotogna: are you using the keys labels and descriptions? if yes, then we should use ObjectUtils, what do you think?

server/modules/category/manager/categoryManager.js
minotogna: do you remember that when transactions fails in Promise.all the exception is not catched? I would avoid promise.all

server/modules/category/manager/categoryManager.js
minotogna: my comment as before

server/modules/category/service/categoryImportJob.js
minotogna: ok right, but it still fetches after delete

fefrain: I'll do it... now ;)

common/survey/categoryItem.js
fefrain: yes, I'll remove them and use ObjectUtils instead

server/modules/category/service/categoryImportJob.js
```diff
@@ -0,0 +1,202 @@
...
+              [CategoryItem.props.code]: codeLevel,
+              [ObjectUtils.keysProps.labels]: labels[levelName],
+              [ObjectUtils.keysProps.descriptions]: descriptions[levelName],
+              [CategoryItem.props.extra]: extra,
+            }
+          )
+        }
+      }
+    }
+    this.incrementProcessedItems()
```
minotogna: extra are still not imported. will you fix this in the next branch or this one?

fefrain: in the next branch

PR #331, opened 2019-09-04
User invite when non published

sgiaccio:


common/auth/authGroups.js
```diff
@@ -56,14 +56,24 @@ R.pipe(
...
 )
 
+const getSurveyAdminGroup = R.pipe(
+  getAuthGroups,
+  R.find(g => g.name === groupNames.surveyAdmin)
+)
+
+const isSurveyAdminGroup = (group, surveyInfo) =>
+  getUuid(getSurveyAdminGroup(surveyInfo)) === getUuid(group)
+
```
minotogna: I suspect getSurveyAdminGroup and isSurveyAdminGroup are properties of survey, and they would become respectively Survey.getAuthGroupAdmin and Survey.isAuthGroupAdmin.

I have also noticed that Authorizer.isSystemAdmin is a property of user; it would become User.isSystemAdmin.

what do you think?

sgiaccio: Yes, refactored

PR #332, opened 2019-09-04
Category import summary UI

fefrain:


common/objectUtils.js
```diff
@@ -62,8 +60,18 @@ const setInPath = (pathArray = [], value = '') => obj => {
   return obj
 }
 
+const clean = obj => {
+  for (const [key, value] of Object.entries(obj))
+    if (value === null || value === undefined || R.isEmpty(obj))
+      delete obj[key]
+
+  return obj
+}
```
minotogna: why this? The problem should be fixed where the the object is created. don't you think so?

common/survey/categoryItem.js
```diff
@@ -20,7 +20,7 @@ const newItem = (levelUuid, parentItemUuid = null, props = {}) => ({
   [keys.uuid]: uuidv4(),
   [keys.levelUuid]: levelUuid,
   [keys.parentUuid]: parentItemUuid,
-  [keys.props]: props,
+  [keys.props]: ObjectUtils.clean(props),
```
minotogna: where are you creating an item with null values in props? the issue is there..what do you think?

server/modules/category/service/categoryImportCSVParser.js
```diff
@@ -62,7 +62,16 @@ const createImportSummary = async (filePath) => {
 
       const level = getOrCreateLevel(columnName, columnType)
 
-      acc[columnName] = CategoryImportSummary.newColumn(columnType, level.name, level.index)
+      const extraDataType = columnType === CategoryImportSummary.columnTypes.extra
+        ? CategoryImportSummary.columnDataTypes.text
+        : null
+
+      const columnProp = columnProps[columnType]
+      const language = columnProp && columnProp.lang
```
minotogna: why shouldn't you have a columProp here?

webapp/loggedin/surveyViews/categoryEdit/components/categoryEditView.scss
```diff
@@ -83,13 +83,18 @@ $borderColor: $greyBorder;
...
-    top: 0px;
+    right: 0;
+    top: 0;
   }
 
   .btn-delete {
     justify-self: center;
   }
 
+  > div {
```
minotogna: why not to add the ```.ellipsis``` class to the div instead of adding the same rule here?

common/survey/categoryItem.js
fefrain: this just avoids to have even empty items in the props, like "descriptions" or "extra"

server/modules/category/service/categoryImportCSVParser.js
fefrain: in the case of "extra" columnType

webapp/loggedin/surveyViews/categoryEdit/components/categoryEditView.scss
fefrain: good to know it exists...!

common/objectUtils.js
fefrain: it simplifies the code that generates the object itself

PR #335, opened 2019-09-06
Email

sgiaccio:


common/i18n/resources/en.js
```diff
@@ -455,32 +455,55 @@ module.exports = {
...
     },
     surveyEditor: {
-      label: 'Survey editors',
+      label: 'Survey editor',
+      label_plural: 'Survey editors',
       description: 'Can edit survey, records, invite users',
     },
     dataEditor: {
-      label: 'Data editors',
+      label: 'Survey editor',
```
fefrain: it should be 'Data editor'

common/i18n/resources/en.js
```diff
@@ -455,32 +455,55 @@ module.exports = {
...
     dataAnalyst: {
-      label: 'Data analysts',
+      label: 'Data analyst',
+      label_plural: 'Data analysts',
       description: 'Can edit records in data analysis step',
     },
     surveyGuest: {
       label: 'System guest',
+      label_plural: 'System guests',
       description: 'Can view records',
```
minotogna: surveyGuest label should be 'Survey guest' not 'System guest'.

server/modules/user/api/userApi.js
```diff
@@ -26,7 +26,8 @@ module.exports.init = app => {
         throw new SystemError('invalidUser')
       }
 
-      await UserService.inviteUser(user, surveyId, email, groupUuid)
+      const serverUrl = `${req.protocol}://${req.headers.host}`
```
minotogna: Should we add a method to Request? e.g. ```Request.getServerUrl(req)``` ?
I would also add the port

server/modules/user/service/userService.js
```diff
@@ -27,12 +30,16 @@ const inviteUser = async (user, surveyId, email, groupUuid) => {
   // If the survey is not published, only survey admins and system admins can be invited
   const survey = await SurveyManager.fetchSurveyById(surveyId)
   const surveyInfo = Survey.getSurveyInfo(survey)
+  const surveyLabel = Survey.getLabel(surveyInfo)
```
minotogna: if you don't pass the lang, then you will see the name. I think it's better to pass the lang in order to show the label if there is.

server/system/aws.js
```diff
@@ -14,12 +14,14 @@ const _sendAwsRequest = request =>
     })
   })
 
-const inviteUser = email => {
+const inviteUser = async (email, temporaryPassword) => {
```
fefrain: you should add "await" before _sendAwsRequest or remove the "async" keyword

server/system/aws.js
```diff
@@ -49,7 +51,39 @@ const updateEmail = (oldEmail, newEmail) => {
...
+    ReplyToAddresses: [from],
+  }
+
+  // Create the promise and SES service object
+  const sendEmailPromise = new aws.SES({
+    apiVersion: '2010-12-01',
+    region: process.env.COGNITO_REGION,
+  }).sendEmail(params).promise()
+
+  return sendEmailPromise
```
fefrain: sendEmailPromise is a redundant variable

webapp/login/components/useResetPasswordFormState.js
```diff
@@ -27,7 +27,7 @@ export const useResetPasswordFormState = props => {
   const validateObj = async obj => await validate(
     obj,
     {
-      'password': [validateRequired, validatePassword, validatePasswordStrength],
+      'password': [validateRequired('Password is required'), validatePassword, validatePasswordStrength],
```
fefrain: should this be a key? something like 'validationErrors.user.passwordRequired'?

common/i18n/resources/en.js
sgiaccio: Right, changed

server/modules/user/service/userService.js
sgiaccio: Ok, added the lang

common/i18n/resources/en.js
sgiaccio: Right

server/system/aws.js
sgiaccio: Removed the asyncs

server/system/aws.js
sgiaccio: Removed the variable

server/modules/user/api/userApi.js
sgiaccio: Ok added. Chrome, Firefox and Safari seem to add the port number to the request header

webapp/login/components/useResetPasswordFormState.js
sgiaccio: Yes, added keys

common/auth/authorizer.js
```diff
@@ -8,30 +8,23 @@ const AuthGroups = require('./authGroups')
...
 
-const isSurveyAdmin = (user, surveyInfo) =>
+const _isSurveyAdmin = (user, surveyInfo) =>
   Survey.isAuthGroupAdmin(getSurveyUserGroup(user, surveyInfo))(surveyInfo)
 
 // ======
 // ====== Survey
 // ======
 
+const _getNonSystemAdminSurveyUserGroup = (user, surveyInfo) =>
```
minotogna: what about adding a boolean param e.g.:  ```getSurveyUserGroup(user, surveyInfo, includeSystemAdmin = true)```. so when you call it from below, you pass false ?
(Instead of adding this method)

common/auth/authorizer.js
```diff
@@ -89,57 +80,34 @@ const canInviteUsers = hasSurveyPermission(permissions.userInvite)
...
+  return User.isSystemAdmin(user) || (
+    !!getSurveyUserGroup(user, surveyInfo) && !!getSurveyUserGroup(userToView, surveyInfo)
+  )
 }
 
 // EDIT
-const isSurveyUser = (surveyInfo, user) => AuthGroups.getAuthGroups(user).some(
-  g => AuthGroups.getSurveyUuid(g) === Survey.getUuid(surveyInfo)
-)
+const _hasUserAccess = (user, surveyInfo, userToUpdate) =>
```
minotogna: i would call this ```_hasUserEditAccess```. what do you think?

server/utils/request.js
```diff
@@ -2,6 +2,8 @@ const R = require('ramda')
 
 const SystemError = require('../../server/utils/systemError')
 
+const getServerUrl = req => `${req.protocol}://${req.headers.host}`
```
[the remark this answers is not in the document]
minotogna: actually it should be ```${req.protocol}://${req.get('host')}${req.originalUrl}```

common/auth/authorizer.js
sgiaccio: Ok

common/auth/authorizer.js
```diff
@@ -89,57 +80,34 @@ const canInviteUsers = hasSurveyPermission(permissions.userInvite)
 
 // READ
 const canViewUser = (user, surveyInfo, userToView) => {
-  if (User.isSystemAdmin(user)) {
-    return true
-  }
-
-  return !!getSurveyUserGroup(user, surveyInfo) && !!getSurveyUserGroup(userToView, surveyInfo)
+  return User.isSystemAdmin(user) || (
+    !!getSurveyUserGroup(user, surveyInfo) && !!getSurveyUserGroup(userToView, surveyInfo)
```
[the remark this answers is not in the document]
sgiaccio: Right, fixed

common/auth/authorizer.js
sgiaccio: Ok

server/utils/request.js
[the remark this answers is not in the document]
sgiaccio: But then would be http://<host>/api/survey/<surveyId>/users/invite in our case. Is this what we want?

package.json
```diff
@@ -72,7 +72,9 @@
     "fast-csv": "^3.4.0",
     "fast-xml-parser": "^3.12.20",
     "file-saver": "^2.0.2",
+    "generate-password": "^1.4.2",
     "i18next": "^17.0.11",
+    "i18next-node-fs-backend": "^2.1.3",
```
minotogna: do we still need this?

server/utils/mailer.js
```diff
@@ -0,0 +1,15 @@
+const aws = require('../system/aws')
+
+const i18n = require('../system/i18n')
+
+const sendEmail = async (to, msgKey, msgParams = {}, lang) => {
+  i18n.translate(async t => {
```
minotogna: why can't you simply use ```i18nFactory.createI18nPromise``` ? do we really need a new file?

server/utils/request.js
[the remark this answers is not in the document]
minotogna: no, you're right

package.json
sgiaccio: See answer below, needed if we load the language at server startup

server/utils/mailer.js
sgiaccio: Using i18nFactory

package.json
sgiaccio: Switched so i18nFactory so not needed

PR #337, opened 2019-09-10
Collect import sampling points

fefrain:


server/job/job.js
```diff
@@ -261,11 +261,21 @@ class Job {
...
   }
 
+  //TODO replace it with user getter
   getUser () {
+    return this.user
+  }
+
+  get user () {
     return this.getContextProp(Job.keysContext.user)
   }
```
minotogna: I'm not sure i like those TODOs. we should use the old way or take care of this after this pull request. what do you think?

server/utils/file/csvParser.js
```diff
@@ -83,6 +83,7 @@ class CSVParser {
 
     if (this._headers === null) {
       this._headers = data
+      console.log('headers', this._headers, this.headers)
```
minotogna: we don't want to use console.log anymore. please use Logger

server/utils/file/csvReader.js
```diff
@@ -60,14 +58,28 @@ const createReader = (filePath, onHeaders, onRow, onTotalChange) => {
       .on('error', reject)
   })
 
+  const calculateSize = () => new Promise((resolve, reject) => {
```
minotogna: do you really need this?

server/utils/file/fileZip.js
```diff
@@ -48,7 +48,10 @@ class FileZip {
 
   async getEntryStream (entryName) {
     return new Promise(resolve =>
-      this.streamZip.stream(entryName, (err, stm) => resolve(stm))
+      this.streamZip.stream(entryName, (err, stm) =>
+        err
+          ? resolve(null)
+          : resolve(stm))
```
minotogna: should it be reject if there's err ?

fefrain: you are right, we should check first if the entry exists, if it doesn't exist we can resolve to null, otherwise reject in case of errors

server/utils/file/csvReader.js
fefrain: yes, it's used in taxonomiesImportJob to better estimate the total items to be processed

server/utils/file/csvParser.js
fefrain: I'll remove it, even csvParser (used in taxonomyImportJob still) should be replaced by csvReader

server/job/job.js
fefrain: we can do a search and replace after this pull request, it makes the code a bit more readable

minotogna: yes i agree, please add a Trello issue

server/utils/file/csvParser.js
minotogna: yes, let's postpone it though

server/utils/file/csvReader.js
minotogna: it is super fast when using the onTotalChange callback. I would suggest to remove it

fefrain: ok, I'll remove it

server/modules/category/service/categoryImportJob.js
```diff
@@ -17,231 +18,248 @@ const CategoryImportJobParams = require('./categoryImportJobParams')
...
+          }
+        }
+        return accExtraDef
+      },
+      {}
+    )
+  }
 
-    let category = await CategoryManager.deleteLevelsByCategory(user, surveyId, categoryUuid, this.tx)
+  extractItemExtraProps (extra) {
```
minotogna: what is this? a method that returns the argument?

server/modules/category/service/categoryImportJob.js
```diff
@@ -17,231 +18,248 @@ const CategoryImportJobParams = require('./categoryImportJobParams')
...
-    const summary = CategoryImportJobParams.getSummary(this.params)
-
-    this.logDebug('summary', summary)
-
-    const levels = Category.getLevelsArray(category)
-
-    const reader = await CategoryImportCSVParser.createRowsReader(
-      summary,
+    const reader = await CategoryImportCSVParser.createRowsReaderFromStream(
+      await this.createReadStream(),
```
minotogna: why do you create a steam if you have the file path?

server/modules/category/service/categoryImportJob.js
```diff
@@ -17,231 +18,248 @@ const CategoryImportJobParams = require('./categoryImportJobParams')
...
+      this._addError(ValidatorErrorKeys.categoryImport.emptyFile)
+    } else if (this.hasErrors()) {
+      // errors found in csv rows
       this.logDebug(`${Object.keys(this.errors).length} errors found`)
       await this.setStatusFailed()
     } else {
       // no errors found
-      await this.insertItems()
+      // 6. delete unused levels
+      this.category = await CategoryManager.deleteLevelsFromIndex(this.user, this.surveyId, this.category, this.levelIndexDeepest, this.tx)
```
minotogna: why do you need CategoryManager.deleteLevelsFromIndex? would the code be shorter if you just delete all levels and items before importing them? why do you replace  the old ones and then remove the unused levels?

server/modules/category/service/categoryImportJob.js
```diff
@@ -17,231 +18,248 @@ const CategoryImportJobParams = require('./categoryImportJobParams')
...
-    if (this.hasErrors()) {
-      // errors found
+    // 1. initialize summary (get it from params by default)
+    this.summary = await this.getOrCreateSummary()
+
+    this.logDebug('summary', this.summary)
+
+    //skip import if summary is not specified
+    if (!this.summary)
+      return
```
minotogna: why shouldn't you have the summary?

PR #346, opened 2019-09-16
Add scheduler to remove temp files

fefrain:


server/system/schedulers/recordPreviewCleanup.js
```diff
@@ -1,7 +1,7 @@
 const schedule = require('node-schedule')
 
-const SurveyManager = require('../modules/survey/manager/surveyManager')
-const RecordService = require('../modules/record/service/recordService')
+const SurveyManager = require('../../modules/survey/manager/surveyManager')
+const RecordService = require('../../modules/record/service/recordService')
```
minotogna: Can we use jus RecordService here ? (i know it was already there)

server/system/schedulers/tempFilesCleanup.js
```diff
@@ -0,0 +1,60 @@
...
+  try {
+    const files = await fs.readdirSync(tempFolder)
+    for (const file of files) {
+      const filePath = path.join(tempFolder, file)
+      const stat = await fs.statSync(filePath)
+      if (stat.isFile())
+        await visitor(filePath, stat)
+    }
+  } catch (error) {
+    //ignore errors (eg. temp folder not created yet, unable to delete temp file, etc)
```
minotogna: Can you log the error?

server/system/schedulers/tempFilesCleanup.js
```diff
@@ -0,0 +1,60 @@
...
+    // cleanup temp files older than 6 hours
+    await cleanupTempFiles(6)
+  })
+
+const cleanupTempFiles = async (olderThanHours = null) => {
+  Logger.debug('Cleaning up temp files')
+
+  let count = 0
+  const now = new Date()
+
```
minotogna: I don't really see the need to call the function visit temp files with the visitor to that just delete the file. Why not to do everything in this method?

fefrain: ok

server/system/schedulers/tempFilesCleanup.js
fefrain: ok

server/system/schedulers/recordPreviewCleanup.js
fefrain: right

server/system/appCluster.js
```diff
@@ -66,4 +67,5 @@ module.exports = async () => {
   // ====== schedulers
   await RecordPreviewCleanup.init()
   await ExpiredJwtTokensCleanup.init()
+  await TempFilesCleanup.init()
 }
```
sgiaccio: Why don't we put the path to the tempFolder in .env and initialize goth the logger and the TempFilesCleanup with its value?

fefrain: very good idea, thanks!

minotogna: Cool. Please update the read me as well

server/system/environment.js
```diff
@@ -0,0 +1,13 @@
...
+  tempFolder: 'TEMP_FOLDER'
+}
+
+const defaultTempFolder = '/tmp/arena_upload'
+
+const tempFolder = R.propOr(defaultTempFolder, keys.tempFolder)(process.env)
+
+module.exports = {
+  tempFolder
+}
```
minotogna: what about adding to common/processUtils and ENV object that contains all the env variables? this mean all references to process.env should be replace by ```ProcessUtils.ENV.[var]```?

server/system/environment.js
```diff
@@ -0,0 +1,13 @@
+const R = require('ramda')
+
+const keys = {
+  tempFolder: 'TEMP_FOLDER'
+}
+
+const defaultTempFolder = '/tmp/arena_upload'
+
+const tempFolder = R.propOr(defaultTempFolder, keys.tempFolder)(process.env)
```
minotogna: defaultTempFolder can be inline

server/system/environment.js
fefrain: honestly I don't like the "utils" prefix, and I would prefer to have a getter for every environment variable

server/system/environment.js
fefrain: ok

server/system/environment.js
minotogna: why a getter ?

server/utils/mailer.js
```diff
@@ -7,7 +8,7 @@ const sendEmail = async (to, msgKey, msgParams = {}, lang) => {
   const mailSubject = i18n.t(`${msgKey}.subject`)
   const msgBody = i18n.t(`${msgKey}.body`, msgParams)
 
-  await aws.sendEmail(process.env.ADMIN_EMAIL, to, mailSubject, msgBody)
+  await aws.sendEmail(ProcessUtils.ENV.adminEmail, to, mailSubject, msgBody)
```
minotogna: do you mean ProcessEnv.adminEmail?

minotogna: I also didn't see that we need to specify ADMIN_EMAIL in .env @sgiaccio and neither we were notified. @fefrain could you add it to the README, thanks.

server/system/processEnv.js
```diff
@@ -0,0 +1,25 @@
+const ProcessUtils = require('../../common/processUtils')
+
```
minotogna: to me ProcessEnv should go under server/utils. It's not used only in system, but rather into several server modules

server/utils/mailer.js
fefrain: ok

server/system/processEnv.js
fefrain: ok

webpack.config.js
```diff
@@ -9,15 +9,17 @@ const BundleAnalyzerPlugin = require('webpack-bundle-analyzer').BundleAnalyzerPl
...
+const ProcessUtils = require('./common/processUtils')
+
 const mode = {
   development: 'development',
   production: 'production'
 }
 
-const prodBuild = process.env.NODE_ENV === mode.production
-const buildReport = process.env.BUILD_REPORT === 'true'
+const prodBuild = ProcessUtils.ENV.nodeEnv === mode.production
```
minotogna: can we add ProcessUtils.envProduction ? (internal mode object should be removed)

PR #348, opened 2019-09-16
Geo service

fefrain:


common/geo/srs.js
```diff
@@ -0,0 +1,41 @@
...
+}
+
+const newSrs = (code, name, wkt) => ({
+  [keys.code]: code,
+  [keys.name]: name,
+  [keys.wkt]: wkt,
+})
+
+//EPSG:4326 WGS84 Lat Lon Spatial Reference System
+const latLonSrs = newSrs('4326', 'GCS WGS 1984', 'GEOGCS["WGS 84",\n' +
```
minotogna: I think it might be more readable with new js strings `` so you won't need the + and the \n

common/geo/geoUtils.js
```diff
@@ -0,0 +1,79 @@
...
+ */
+const srs = R.pipe(
+  R.concat(projected.ProjectedCoordinateSystems),
+  R.concat(geographic.GeographicCoordinateSystems),
+  R.map(item => Srs.newSrs(item.wkid.toString(), formatName(item.name), item.wkt)),
+  R.sortBy(R.prop(Srs.keys.name))
+)([])
+
+const srsByCode = ObjectUtils.toIndexedObj(srs, Srs.keys.code)
+
```
minotogna: should all srs related code go to srs.js ?

common/survey/survey.js
```diff
@@ -11,13 +11,15 @@ const SurveyDefaults = require('./_survey/surveyDefaults')
...
+const Srs = require('../geo/srs')
+
 const newSurvey = (ownerUuid, name, label, lang, collectUri = null) => ({
   uuid: uuidv4(),
   props: {
     name,
     labels: { [lang]: label },
     languages: [lang],
-    srs: [{ code: '4326', name: 'GCS WGS 1984' }], //EPSG:4326 WGS84 Lat Lon Spatial Reference System,
+    srs: [Srs.latLonSrs],
```
minotogna: you are adding all the WKT to survey. is that what we want ?

webapp/loggedin/modules/designer/surveyInfo/components/srsEditor.js
```diff
@@ -21,7 +21,7 @@ const SrsEditor = props => {
     <InputChips
       itemsLookupFunction={srsLookupFunction}
       itemKeyProp="code"
-      itemLabelProp="name"
+      itemLabelFunction={srs => `${srs.name} (EPSG:${srs.code})`}
```
minotogna: should we use SRS.getName and SRS.getCode ?

common/geo/srs.js
fefrain: right

common/survey/survey.js
fefrain: right, I'll change it

common/geo/geoUtils.js
minotogna: what I mean is that srs (maybe rename to srsArray ?), srsByCode, getSrsByCode should be moved to srs.js ?

fefrain: srs.js is used by the client too... I'll rename the internal constant into srsArray anyway

webapp/loggedin/modules/designer/surveyInfo/components/srsEditor.js
fefrain: right

common/geo/geoUtils.js
minotogna: can you write a comment then?

fefrain: ok

PR #363, opened 2019-09-23
SurveyFormView: show error badge orange when there's a warning

fefrain:


common/record/_recordValidator/attributeValidator.js
```diff
@@ -42,9 +42,7 @@ const _validateNodeValidations = (survey, record, nodeDef) => async (propName, n
     if (!valid) {
       const messages = _getCustomValidationMessages(survey, expression)
 
-      errorMessage = ValidationResult.newInstance(
-        Validation.keys.customErrorMessageKey,
-        null,
+      errorMessage = ValidationResult.newInstanceCustom(
```
minotogna: why a custom constructor? I would have left as it was. otherwise for every exception we need to add a specific constructor

webapp/commonComponents/errorBadge.scss
```diff
@@ -14,4 +18,13 @@
     font-weight: bold;
     color: $red;
   }
+
+  &.warning {
+    background-color: transparent;
```
minotogna: isn't is inherited from ```error-badge-inverse```?

webapp/commonComponents/validationTooltip.js
```diff
@@ -11,7 +11,13 @@ export default ({ validation, className, showKeys, children }) => {
   const i18n = useI18n()
 
   const isValid = Validation.isValid(validation)
-  const type = isValid ? '' : 'error'
+
+  const type = isValid
+    ? ''
+    : Validation.hasWarnings(validation) || Validation.hasWarningsInFields(validation)
```
minotogna: ```Validation.hasWarnings(validation) || Validation.hasWarningsInFields(validation) ``` is used several times. can we have a method  ```Validation.isWarning```?

webapp/commonComponents/errorBadge.js
```diff
@@ -11,13 +13,19 @@ import ValidationTooltip from './validationTooltip'
...
   } = props
 
   const i18n = useI18n()
   const invalid = !Validation.isValid(validation)
 
+  // when there are warnings add 'warning' class to className
+  const className = R.when(
+    R.always(Validation.hasWarnings(validation) || Validation.hasWarningsInFields(validation)),
+    R.concat('warning ')
+  )(classNameProps)
```
minotogna: warning should go after

fefrain: right

webapp/commonComponents/validationTooltip.js
fefrain: ok

common/record/_recordValidator/attributeValidator.js
fefrain: I'll remove it

webapp/commonComponents/errorBadge.scss
fefrain: apparently the definition .error-badge.warning is more specific than .error-badge-inverse.warning and overrides it

PR #367, opened 2019-09-24
Category validation persist validation

fefrain:


common/survey/_survey/surveyCategories.js
```diff
@@ -11,6 +13,17 @@ const getCategories = R.pipe(
 const getCategoriesArray = R.pipe(
   getCategories,
   R.values,
+  //sort by name
+  R.sort((category1, category2) => {
+    const name1 = Category.getName(category1)
+    const name2 = Category.getName(category2)
+
```
minotogna: should they be both converted to upper case or lower case?

common/survey/category.js
```diff
@@ -45,19 +45,15 @@ const newCategory = (props = {}) => {
...
-
-  return {
-    [CategoryLevel.keys.uuid]: uuidv4(),
-    [CategoryLevel.keys.categoryUuid]: ObjectUtils.getUuid(category),
-    [CategoryLevel.keys.index]: index,
-    [CategoryLevel.keys.props]: {
-      [CategoryLevel.props.name]: 'level_' + (index + 1),
-      ...props
-    }
+const newLevel = (category, props = {}, index = getLevelsArray(category).length) => ({
```
minotogna: why not to use ```Object.keys(Category.getLevels(category)).length```

server/modules/survey/service/publish/jobs/categoriesValidationJob.js
```diff
@@ -12,15 +12,16 @@ class CategoriesValidationJob extends Job {
   }
 
   async execute (tx) {
-    const categories = await CategoryManager.fetchCategoriesBySurveyId(this.surveyId, true, false, tx)
+    const categories = await CategoryManager.fetchCategoriesAndLevelsBySurveyId(this.surveyId, true, true, tx)
 
-    this.total = categories.length
+    const categoriesArr = Object.values(categories)
```
minotogna: why don't you iterate with Object.entries?

common/survey/category.js
fefrain: right

server/modules/survey/service/publish/jobs/categoriesValidationJob.js
fefrain: you need only the values, not the keys (uuids) of categories

minotogna: yes, but you first get the values and then iterate. what i meant you can get the entries only. then iterate over them as well as get the length

PR #368, opened 2019-09-25
Fixed category tests

fefrain:


server/modules/record/manager/recordManager.js
```diff
@@ -40,7 +60,8 @@ const fetchRecordAndNodesByUuid = async (surveyId, recordUuid, draft = true, cli
 
 module.exports = {
   // ==== CREATE
-  insertRecord: RecordRepository.insertRecord,
+  insertRecord,
+  insertAndInitRecord,
```
minotogna: I doubt we need those 2 methods. can you replicate the thread logic in the tests?

fefrain: insertRecord is needed because it inserts even a record in ActivityLog.
I'll move insertAndInitRecord in the /test/it/utils/recordUtils.js file.

PR #369, opened 2019-09-25
Category published

fefrain:


server/modules/category/repository/categoryRepository.js
```diff
@@ -194,6 +206,12 @@ const updateCategoryValidation = async (surveyId, categoryUuid, validation, clie
     [validation, categoryUuid]
   )
 
+const markCategoriesPublished = async (surveyId, client = db) =>
```
minotogna: I would call this method ```markCategoriesPublishedBySurveyId``` what do you think?

server/modules/category/repository/categoryRepository.js
```diff
@@ -107,6 +108,17 @@ const fetchCategoryAndLevelsByUuid = async (surveyId, categoryUuid, draft = fals
   return R.pipe(R.values, R.head)(categories)
 }
 
+const isCategoryPublished = async (surveyId, categoryUuid, client = db) => {
```
minotogna: I wouldn't create a method just to check if it's published. you can use fetchCategoryAndLevlesByCategoryUuid and use Category.isPublished.

server/modules/category/repository/categoryRepository.js
fefrain: Is there any other markCategoriesPublished we should distinguish from?

server/modules/category/repository/categoryRepository.js
fefrain: In this way you avoid to fetch all the category, but I'll change it

minotogna: in this way, if we decide to change how we handle a published category, you need to update the code twice: in the Category.isPublished and in this method. it's about consistency. Fetching 1 category with levels is a very fast operation

server/modules/category/repository/categoryRepository.js
minotogna: just to be consistent with other methods in the same module. whatever works in the end, your call

server/modules/category/repository/categoryRepository.js
fefrain: right, I'll change it

server/modules/category/repository/categoryRepository.js
fefrain: ok

PR #386, opened 2019-09-30
associate cycle to record on creation

fefrain:


common/record/_record/recordKeys.js
```diff
@@ -1,15 +1,16 @@
...
-  ownerUuid: 'ownerUuid',
   ownerName: 'ownerName',
-  step: 'step',
+  ownerUuid: 'ownerUuid',
   preview: 'preview',
-  dateCreated: ObjectUtils.keys.dateCreated,
-  dateModified: ObjectUtils.keys.dateModified,
+  step: 'step',
+  surveyUuid: 'surveyUuid',
+  uuid: 'uuid',
```
minotogna: can you use ObjectUtils.keys.uuid?

webapp/loggedin/surveyViews/record/actions.js
```diff
@@ -54,8 +54,9 @@ export const createRecord = (history, preview = false) => async (dispatch, getSt
   const state = getState()
   const user = AppState.getUser(state)
   const surveyId = SurveyState.getSurveyId(state)
+  const cycle = User.getPrefSurveyCycle(surveyId)(user)
```
minotogna: please, use ```SurveyState.getSurveyCycleKey```

common/record/_record/recordKeys.js
fefrain: sure

webapp/loggedin/surveyViews/record/actions.js
fefrain: since we have to get user and surveyId from the state already, it's better to use directly the User function, isn't it?

minotogna: nope.. I would prefer one common place to get the current cycle

PR #403, opened 2019-10-02
Reset DataQueryState on survey cycle change

fefrain:


webapp/loggedin/modules/data/dataVis/dataQuery/dataQueryState.js
```diff
@@ -144,6 +144,8 @@ export const assocTableDataRecordNodeValidations = (recordUuid, recordValid) =>
     data => R.assocPath([keys.table, tableKeys.data], data)(state)
   )(state)
 
+export const resetTable = R.assoc(keys.table, {})
```
minotogna: you need to reset all state. 
this method would be ```resetDataQueryState```

fefrain: right

PR #405, opened 2019-10-02
SurveyPublishJob: delete records that are associated to removed cycles

fefrain:


server/modules/survey/service/publish/jobs/recordsRemovedCyclesDeleteJob.js
```diff
@@ -0,0 +1,49 @@
...
+const SurveyManager = require('../../../../survey/manager/surveyManager')
+
+class RecordsRemovedCyclesDeleteJob extends Job {
+
+  constructor (params) {
+    super(RecordsRemovedCyclesDeleteJob.type, params)
+  }
+
+  async execute (tx) {
+    this.total = 2
```
minotogna: what about ```this.total = 1 + cycleKeysDeleted.length```?

fefrain: uhm, the calculation of cycleKeysDeleted requires some time, and it's not even correct to add the entire lenght of cycleKeysDeleted to the total (the deletion is done with a single query)

minotogna: ok!

PR #406, opened 2019-10-03
SurveyPublishJob: UserPrefs set to '0' cycle pref when cycle is among…

fefrain:
… deleted ones


server/modules/user/repository/userRepository.js
```diff
@@ -126,6 +127,19 @@ const deleteUsersPrefsSurvey = async (surveyId, client = db) => {
 `)
 }
 
+/**
+ * Sets survey cycle user pref to Survey.cycleOneKey if the preferred cycle is among the specified (deleted) ones
+ */
+const deleteUsersPrefsCycleDeleted = async (surveyId, cycleKeysDeleted, client = db) => {
```
minotogna: Maybe this method should just be ```deleteUsersPrefsSurveyCycle = async (surveyId, cycles, client = db) ```

fefrain: or even better, resetUsersPrefsSurveyCycle

server/modules/survey/service/publish/jobs/userPrefsCycleResetJob.js
```diff
@@ -0,0 +1,36 @@
+const R = require('ramda')
+
+const Job = require('../../../../../job/job')
+
+const UserManager = require('../../../../user/manager/userManager')
+const SurveyPublishJobUtils = require('../surveyPublishJobUtils')
+
+class UserPrefsCycleResetJob extends Job {
```
minotogna: what about one job called ```CyclesDeletedCheckJob``` that removes the records and users prefs?

fefrain: good idea

PR #409, opened 2019-10-04
Record check job

fefrain:


common/i18n/resources/en.js
```diff
@@ -394,6 +394,14 @@ module.exports = {
       emptyFile: 'The file you are trying to import is empty'
     },
 
+    expressions: {
+      expressionInvalid: "Invalid expression: {{details}}",
+      cannotGetChildOfAttribute: 'cannot get child node {{childName}} of attribute {{parentName}}',
+      unableToFindNodeChild: 'unable to find child node: {{name}}',
+      unableToFindNodeParent: 'unable to find parent node: {{name}}',
+      unableToFindNodeSibling: 'unable to find sibling node: {{name}}',
```
minotogna: Please use first letter Upper case

PR #412, opened 2019-10-07
Hide app loader on nodesUpdate

sgiaccio:


webapp/loggedin/surveyViews/record/actions.js
```diff
@@ -25,8 +25,10 @@ export const nodesUpdate = 'survey/record/node/update'
 export const nodeDelete = 'survey/record/node/delete'
 export const validationsUpdate = 'survey/record/validation/update'
 
-export const recordNodesUpdate = nodes => dispatch =>
+export const recordNodesUpdate = nodes => dispatch => {
   dispatch({ type: nodesUpdate, nodes })
+  dispatch(hideAppLoader())
```
minotogna: this will be dispatched at every nodes update. can we add a check to dispatch this only when record has no nodes?

sgiaccio: Ok

PR #424, opened 2019-10-10
ProcessingChain: handle creation from list view

fefrain:


server/modules/activityLog/activityLogger.js
```diff
@@ -48,6 +48,7 @@ const type = {
   userRemove: 'userRemove',
 
   // analysis
+  processingChainInsert: 'processingChainInsert',
```
minotogna: should we use processingChainCreate? it's more consistent with the other activity logs.. what do you think?

server/modules/analysis/manager/processingChainManager.js
```diff
@@ -1,14 +1,28 @@
...
 const ActivityLog = require('../../activityLog/activityLogger')
+
+const ProcessingChain = require('../../../../common/analysis/processingChain')
+
 const ProcessingChainRepository = require('../repository/processingChainRepository')
 
+const createChain = async (user, surveyId, cycle, client = db) =>
+  await client.tx(async t => {
+    const processingChain = await ProcessingChainRepository.insertChain(surveyId, cycle, t)
+    await ActivityLog.log(user, surveyId, ActivityLog.type.processingChainInsert, { processingChain }, t)
```
minotogna: I think the repository can return the uuid directly and save in the history the uuid . is there a reason to save all the processing chai in the activity log?

webapp/loggedin/modules/analysis/processingChains/processingChainsView.js
```diff
@@ -11,15 +11,16 @@ import TableView from '../../../tableViews/tableView'
...
 import * as SurveyState from '../../../../survey/surveyState'
 
 import { reloadListItems } from '../../../tableViews/actions'
-import { appModuleUri, analysisModules } from '../../../appModules'
+import { createProcessingChain, navigateToProcessingChainView } from './actions'
 
 const processingChainsModule = 'processing-chains'
 
-const ProcessingChainsView = ({ surveyCycleKey, reloadListItems, history }) => {
+const ProcessingChainsView = ({ surveyCycleKey, reloadListItems, history, createProcessingChain, navigateToProcessingChainView }) => {
```
minotogna: it would be more readable if you use props:
```
const ProcessingChainsView = props => {
    const { 
        surveyCycleKey, history, 
        reloadListItems,  createProcessingChain, navigateToProcessingChainView
     } = props
....
```

server/modules/activityLog/activityLogger.js
fefrain: ok, right

server/modules/analysis/manager/processingChainManager.js
fefrain: if you want to recreate the same processing chain (with cycle, props, date created/modified etc) you have to store it entirely

webapp/loggedin/modules/analysis/processingChains/processingChainsView.js
fefrain: ok

PR #439, opened 2019-10-16
ActivityLog: record userInvitationAccept log when use accepts invitation

fefrain:


server/modules/user/manager/userManager.js
```diff
@@ -98,6 +98,18 @@ const updateUserPrefs = async user => ({
   [User.keys.authGroups]: await AuthGroupRepository.fetchUserGroups(User.getUuid(user))
 })
 
+const acceptInvitation = async (user, name, client = db) =>
+  await client.tx(async t => {
+    const userUpdated = await UserRepository.updateUsername(user, name, t)
+    const groups = await AuthGroupRepository.fetchUserGroups(User.getUuid(user), t)
```
minotogna: when user accept an invitation, he has only one group associated to that survey.

server/modules/user/service/userService.js
```diff
@@ -126,15 +126,16 @@ const updateUser = async (user, surveyId, userUuid, name, email, groupUuid, file
...
     throw new UnauthorizedError(User.getName(user))
-  }
 
-  const userToUpdate = await UserManager.fetchUserByUuid(userUuid)
-  await aws.updateUser(User.getEmail(userToUpdate), null, name)
-  await UserManager.updateUsername(user, name)
+  await client.tx(async t => {
+    const userToUpdate = await UserManager.acceptInvitation(user, name, t)
+    // update user name in aws
+    await aws.updateUser(User.getEmail(userToUpdate), null, name)
```
minotogna: what if during aws.updateUser something goes wrong? better to surround this with try catch so that the transaction gets rollbacked ?

server/modules/auth/repository/authGroupRepository.js
```diff
@@ -61,12 +61,16 @@ const fetchSurveyGroups = async (surveyId, client = db) =>
 
 const fetchUserGroups = async (userUuid, client = db) =>
   await client.map(`
-    SELECT auth_group.*
-    FROM auth_group_user, auth_group
+    SELECT g.*, s.id as survey_id
+    FROM auth_group_user gu
+      JOIN auth_group g
+        ON g.uuid = gu.group_uuid
+      LEFT OUTER JOIN survey s
```
minotogna: you don't need left outer join, but did you really need to update this query?

server/modules/user/manager/userManager.js
fefrain: can it be that he is invited to more than one survey and when he logs in the first time he has multiple groups associated?

minotogna: no, the invitation happens only for one survey at the time. with the code you wrote, your are logging all groups, even for other surveys .

server/modules/auth/repository/authGroupRepository.js
fefrain: left outer join is not needed, you are right. I've changed the query to get the survey_id (used to log the userInviteAccept activity)

fefrain: no wait, the left outer join it's needed to fetch user groups associated to the systemAdmin user

minotogna: correct, you're right. but still, do you need to update this query ?

fefrain: Should I create another one for this?

fefrain: do we need another fetch function for this? how can I get the surveyId required to log the activities? with a SurveyRepository.fetchSurveyByUuid ?

minotogna: let's just remove the accept invitation activity log

PR #442, opened 2019-10-17
EntityKeyValidator: fixed bug, nodes not validated on sibling entity delete

fefrain:
…delete

### DIFF
--- a/core/record/_recordValidator/attributeKeyValidator.js
+++ b/core/record/_recordValidator/attributeKeyValidator.js
@@ -3,7 +3,9 @@ const R = require('ramda')
 const Survey = require('../../survey/survey')
 const NodeDef = require('../../survey/nodeDef')
 const Record = require('../record')
+const Node = require('../node')
 const Validation = require('../../validation/validation')
+const ObjectUtils = require('../../objectUtils')
 
 const validateAttributeKey = (survey, record, attributeDef) => async (propName, node) => {
   const nodeDefParent = Survey.getNodeDefParent(attributeDef)(survey)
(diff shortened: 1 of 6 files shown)


core/record/_recordValidator/attributeKeyValidator.js
```diff
@@ -19,19 +18,13 @@ const validateAttributeKey = (survey, record, attributeDef) => async (propName,
 
 const _isEntityDuplicate = (survey, record, entity) => {
   // 1. get sibling entities
-  const parentNode = Record.getParentNode(entity)(record)
-  const siblingEntitiesAndSelf = Record.getNodeChildrenByDefUuid(parentNode, Node.getNodeDefUuid(entity))(record)
-  const siblingEntities = R.reject(R.propEq(Node.keys.uuid, Node.getUuid(entity)), siblingEntitiesAndSelf)
+  const siblingEntities = Record.getNodeSiblings(entity)(record)
```
minotogna: does this return all siblings or only sibling entities?

fefrain: the node passed to getNodeSiblings is an entity -> its siblings are entities too

minotogna: why? a tree, inside a plot, cannot a have sibling like plot no? plot location?

fefrain: these are not siblings but children...

fefrain: you are right, I'll add a filter function then

PR #443, opened 2019-10-17
CSVReader: fixed rows not processed synchronously

fefrain:


server/utils/file/csvReader.js
```diff
@@ -5,25 +5,38 @@ const Queue = require('../../../core/queue')
...
+            .then(() => {
+              processNext()
+            })
+            .catch(e => {
+              cancel()
+              reject(e)
+            })
+            .finally(() => {
+              processingRow = false
+            })
```
minotogna: can't this be used with async/await ? what was the problem here?

fefrain: in the end it doesn't change too much, I'll revert to the previous version and add the processingRow handling

PR #444, opened 2019-10-18
Fast-CSV removal

fefrain:


core/objectUtils.js
```diff
@@ -94,6 +94,13 @@ const toIndexedObj = (array, prop) => array.reduce(
 
 const toUuidIndexedObj = R.partialRight(toIndexedObj, [keys.uuid])
 
+const indexByProps = props => array => props.reduce((acc, item, index) => {
```
minotogna: this is not an ObjectUtils method, but rather a csvParser specific issue. this can be handled directly in csvReader processNext. if there are the headers, then just return an object with key values, otherwise the array. what do you think?

webapp/loggedin/surveyViews/taxonomyEdit/components/taxaTableRow.js
```diff
@@ -1,22 +1,20 @@
...
     <>
       <div>{idx + offset + 1}</div>
-      <div>{Taxon.getCode(row)}</div>
-      <div>{Taxon.getFamily(row)}</div>
-      <div>{Taxon.getGenus(row)}</div>
-      <div>{Taxon.getScientificName(row)}</div>
+      <div>{row.code}</div>
+      <div>{row.family}</div>
+      <div>{row.genus}</div>
+      <div>{row.scientific_name}</div>
```
minotogna: why?

core/objectUtils.js
fefrain: good

webapp/loggedin/surveyViews/taxonomyEdit/components/taxaTableRow.js
fefrain: I'll restore the code from the previous version

PR #446, opened 2019-10-21
Remove dependency on react-remarkable

eliask:


webapp/loggedin/appErrors/appErrors.js
```diff
@@ -31,9 +30,7 @@ const AppError = ({ error, closeAppError }) => {
         ERROR {R.path(['response', 'status'], error)}
       </div>
       <div className="message">
-        <Markdown options={{ html: false }}>
-          {i18n.t(key, params)}
-        </Markdown>
+        {i18nMarkdownToText(i18n, key, params)}
```
minotogna: what about have a component in ```app/commonComponents/markdown.js``` to be used like:
```
<Markdown className="message" options={{ html: false }} source={i18n.t(key, params)} />
```

eliask: Sure, could do that. The original dependency was only a 30-50 line component in the first place so I opted for this initially.
So just have almost the same API, then?

minotogna: yes, the reason is that dependency was not updated anymore, and used old react lifecycle class methods,

eliask:
Hopefully this is sufficient, now.

PR #456, opened 2019-10-24
Make user menu the topmost element

eliask:
The user menu should have the highest precedence since it has to be explicitly toggled on.

Currently it gets hidden behind some elements (e.g. the nodeDef editor)

### DIFF
--- a/webapp/style/_vars.scss
+++ b/webapp/style/_vars.scss
@@ -33,6 +33,7 @@ $zIndex20Dropdown: 20;
 $zIndex80AppNotification: 80;
 $zIndex90AppError: 90;
 $zIndex190AppLoader: 190;
+$zIndex900UserMenu: 900;
 
 $shadow1: 0 0 4px rgba($black, .2);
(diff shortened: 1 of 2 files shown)


webapp/style/_vars.scss
```diff
@@ -33,6 +33,7 @@ $zIndex20Dropdown: 20;
 $zIndex80AppNotification: 80;
 $zIndex90AppError: 90;
 $zIndex190AppLoader: 190;
+$zIndex900UserMenu: 900;
```
minotogna: everything should go below appLoader, so maybe this can be ```$zIndex180UserMenu: 180;```

eliask: My use case for having this be the topmost element was to be able to do something in case the loader takes too long / forever.

But the policy then is to disallow operations while something is loading?

minotogna: yes,  we show the appLoader to stop any operation from the user...

PR #467, opened 2019-10-28
DefaultValue - check nodeDef type compatibility

fefrain:


core/record/recordExprValueConverter.js
```diff
@@ -0,0 +1,60 @@
...
+
+  return taxonUuid ? { [Node.valuePropKeys.taxonUuid]: taxonUuid } : null
+}
+
+const _valueExprToValueNodeFns = {
+  [NodeDef.nodeDefType.boolean]: _valueExprToValueBoolean,
+  [NodeDef.nodeDefType.code]: _valueExprToValueCode,
+  [NodeDef.nodeDefType.date]: (survey, record, nodeCtx, valueExpr) => valueExpr,
+  [NodeDef.nodeDefType.decimal]: (survey, record, nodeCtx, valueExpr) => R.ifElse(_isStringOrNumber, Number, R.always(null))(valueExpr),
+  [NodeDef.nodeDefType.integer]: (survey, record, nodeCtx, valueExpr) => R.ifElse(_isStringOrNumber, Number, R.always(null))(valueExpr),
```
minotogna: - you can create a function ```_toPrimitive = (value, Type=String) =>  R.is(String, value) || R.is(Number, value) ? Type(value) : null``` to be used in text, integer, decimal , and instead of _valueExprToCode
- is it ok to return NaN for integer and decimal?

core/record/recordExprValueConverter.js
```diff
@@ -0,0 +1,60 @@
...
+import * as Survey from '@core/survey/survey'
+import * as NodeDef from '@core/survey/nodeDef'
+import * as Record from '@core/record/record'
+import * as Node from '@core/record/node'
+
+const _valueExprToValueBoolean = (survey, record, nodeCtx, valueExpr) =>
+  R.is(String, valueExpr) && R.includes(valueExpr, ['true', 'false'])
+    ? valueExpr
+    : null
+
```
minotogna: is it possible that the expression returns a boolean and not a string?

core/record/recordExprValueConverter.js
fefrain: good idea

core/record/recordExprValueConverter.js
fefrain: it can be, right

core/record/recordExprValueConverter.js
```diff
@@ -0,0 +1,62 @@
...
+import * as Node from '@core/record/node'
+
+const _valueExprToValueBoolean = (survey, record, nodeCtx, valueExpr) =>
+  R.is(Boolean, valueExpr)
+    ? valueExpr
+    : R.is(String, valueExpr) && R.includes(valueExpr, ['true', 'false'])
+    ? valueExpr
+    : null
+
+const _isStringOrNumber = val => R.is(String, val) || R.is(Number, val)
```
minotogna: why having this function where it's used only by ```toPrimitive```?

core/record/recordExprValueConverter.js
```diff
@@ -0,0 +1,62 @@
...
+  [NodeDef.nodeDefType.boolean]: _valueExprToValueBoolean,
+  [NodeDef.nodeDefType.code]: _valueExprToValueCode,
+  [NodeDef.nodeDefType.date]: (survey, record, nodeCtx, valueExpr) => valueExpr,
+  [NodeDef.nodeDefType.decimal]: (survey, record, nodeCtx, valueExpr) => _toPrimitive(valueExpr, Number),
+  [NodeDef.nodeDefType.integer]: (survey, record, nodeCtx, valueExpr) => _toPrimitive(valueExpr, Number),
+  [NodeDef.nodeDefType.taxon]: _valueExprToValueTaxon,
+  [NodeDef.nodeDefType.text]: (survey, record, nodeCtx, valueExpr) => _toPrimitive(valueExpr, String),
+  [NodeDef.nodeDefType.time]: (survey, record, nodeCtx, valueExpr) => valueExpr,
+}
+
```
minotogna: can method names be _toBoolean, _toCode, etc...?

core/record/recordExprValueConverter.js
```diff
@@ -0,0 +1,62 @@
...
+export const valueExprToValueNode = (survey, record, nodeCtx, valueExpr) => {
```
minotogna: method name can be toNodeValue. the object itself is called RecordExprValueConverter, so when you invoke ```RecordExprValueConverter.toNodeValue``` is enough

core/record/recordExprValueConverter.js
fefrain: I'll remove it

core/record/recordExprValueConverter.js
fefrain: ok

core/record/recordExprValueConverter.js
fefrain: right

core/record/recordExprParser.js
```diff
@@ -78,18 +79,27 @@ const bindNode = (survey, record, node) => {
...
+const evalApplicableExpressions = async (survey, record, node, expressions, stopAtFirstFound = false, convertToNodeValues = false) => {
   const applicableExpressions = await _getApplicableExpressions(survey, record, node, expressions, stopAtFirstFound)
 
   return await Promise.all(
-    applicableExpressions.map(async expression => ({
+    applicableExpressions.map(async expression => {
+      const valueEval = await evalNodeQuery(survey, record, node, NodeDefExpression.getExpression(expression))
+      const value =
+        convertToNodeValues
+          ? R.isNil(valueEval) || R.isEmpty(valueEval)
```
eliask: These nested ternary operators feels quite confusing. I would hope that there's a clearer way to express the same logic.
At least I've never seen anyone use them like this before. Maybe sometimes It's nested like `a ? b : c ? d : e` but never ` a ? b ? c : d : e` (though even the latter is not ideal in my books, at least not without parentheses)

eliask: I wonder if `convertToNodeValues` could be made non-conditional, i.e. to always convert or not to convert? I'm not fully clear on the logic yet, but if it were always done in one way then the overall complexity would decrease a little bit.

fefrain: I'll try to make it more readable. It's used in different places and the conversion is not always needed, so we should continue to specify  `convertToNodeValues`.

eliask: But is the change idempotent? It would simplify things in general to omit conditionals in these kinds of cases if there's no functional difference.

It might be a bit of extra work to do, but if there's no difference in the perceptible speed, I think it would be a good tradeoff.

fefrain: ok, I'll move the conversion where it's needed and remove the condition check

PR #468, opened 2019-10-28
Test-created survey: Add warning labels not to use

eliask:
Added warnings to prevent anyone else from using the (somehow broken) test suite created surveys.

### DIFF
--- a/test/it/006activityLogTest.js
+++ b/test/it/006activityLogTest.js
@@ -18,8 +18,8 @@ describe('Activity Log Test', async () => {
 
   it('Activity Log on Survey Creation', async () => {
     const surveyParam = {
-      name: 'DO NOT USE! test_survey_' + uuidv4(),
-      label: 'Test Survey',
+      name: 'do_not_use__test_survey_' + uuidv4(),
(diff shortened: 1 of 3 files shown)


test/it/006activityLogTest.js
```diff
@@ -18,7 +18,7 @@ describe('Activity Log Test', async () => {
 
   it('Activity Log on Survey Creation', async () => {
     const surveyParam = {
-      name: 'test_survey_' + uuidv4(),
+      name: 'DO NOT USE! test_survey_' + uuidv4(),
```
fefrain: I would have used a "normalized" name in this case, in the way we normalize internal names (lowercase, without spaces) because the survey name should be like that.
During tests the surveys are published: we don't validate the survey name on survey publish because it has been already validated on survey creation and on survey edit, but we could check it in theory.

eliask: Ah, I didn't notice that. I guess it's the labels that are meant to be the human-readable names (and the names likes this I guess are mostly hidden). I'll correct that.

PR #471, opened 2019-10-29
Activity log repository

fefrain:


server/modules/activityLog/activityLog.js
```diff
@@ -1,10 +1,6 @@
-const db = require('@server/db/db')
+import * as R from 'ramda'
```
minotogna: I would move this to '@common/activityLog/activityLog.js'

server/modules/activityLog/api/activityLogApi.js
```diff
@@ -0,0 +1,20 @@
+import * as Request from '@server/utils/request'
+import * as AuthMiddleware from '../../auth/authApiMiddleware'
+import * as ActivityLogService from '../service/activityLogService'
+
+export const init = app => {
+  // ==== READ
+
+  app.get(`/survey/:surveyId/activity-log`, AuthMiddleware.requireSurveyEditPermission, async (req, res, next) => {
+    try {
+      const { surveyId, offset } = Request.getParams(req)
```
minotogna: I would pass also limit as rest parameter

fefrain: I didn't pass it to avoid to check that parameter; should it be always required and limited to 100 for example?

server/modules/activityLog/activityLog.js
fefrain: it's used only by the server modules now, I'd move it only if needed.

minotogna: it will be needed in the UI. i would move it in this pull request, otherwise in the next one we'll need to review the same files again

server/modules/activityLog/api/activityLogApi.js
minotogna: you need the limit in the UI to add the placeholder to incrementally fetch the next sequence of items as we discussed earlier, that's why i'm suggesting you add it now

fefrain: but the limit in the query is always the same, I think you will pass different offsets every time... right?

PR #475, opened 2019-10-30
Recent activities filter by user permissions

fefrain:


server/modules/activityLog/api/activityLogApi.js
```diff
@@ -1,18 +1,17 @@
...
 
 import * as Request from '@server/utils/request'
-import * as AuthMiddleware from '../../auth/authApiMiddleware'
 import * as ActivityLogService from '../service/activityLogService'
 
 export const init = app => {
   // ==== READ
 
-  app.get(`/survey/:surveyId/activity-log`, AuthMiddleware.requireSurveyEditPermission, async (req, res, next) => {
+  app.get(`/survey/:surveyId/activity-log`, async (req, res, next) => {
```
minotogna: you need to use ```AuthMiddleware.requireSurveyViewPermission```

test/it/utils/recordUtils.js
```diff
@@ -109,6 +110,25 @@ const getValidationMaxCount = (parentNode, childDef) => R.pipe(
   Validation.getFieldValidation(RecordValidation.keys.maxCount)
 )
 
+const updateNodeAndExpectDependentNodeValueToBe = async (survey, record, sourcePath, sourceValue, dependentPath, dependentExpectedValue) => {
```
minotogna: where do you use this?

server/modules/activityLog/api/activityLogApi.js
fefrain: ok

test/it/utils/recordUtils.js
fefrain: right, I've committed it by mistake

core/auth/authorizer.js
```diff
@@ -59,51 +49,51 @@ const canEditRecord = (user, record) => {
...
 // CREATE
-const canInviteUsers = hasSurveyPermission(permissions.userInvite)
+const canInviteUsers = _hasSurveyPermission(permissions.userInvite)
 
 // READ
 const canViewUser = (user, surveyInfo, userToView) => {
   return User.isSystemAdmin(user) || (
-    !!getSurveyUserGroup(user, surveyInfo, false) &&
-    !!getSurveyUserGroup(userToView, surveyInfo, false)
+    !!_getSurveyUserGroup(user, surveyInfo, false) &&
```
eliask: Are the `!!` necessary here? It just coerces the value to boolean?
Is it because these functions are defined to return booleans rather than the "equivalent" true-y/falsy values?

PR #477, opened 2019-10-31
Expression changes

eliask:
This pull request changes the expression syntax and semantics from "nearly JS" to "more abstract/mathematical (-ish)".

The parser in use is still JSEP, but some of its functionality is currently not used and are wired to give errors, e.g. "this", and "foo.bar".

Function names are constrained to be simple strings and have a different namespace from variables/fields/nodes, so it's possible to say `pow(2, 3) === pow` (where the right side is e.g. a field that has a value of 0).

Previous uses of `eval` were removed to gain full control over operator and "standard library" function semantics.

Introduced a standard library of functions that now consists of just the following: `min`, `max`, `pow`, as well as the unary and binary operators (bitwise operators removed).

Functional dependencies between expressions and nodes currently use the expression parser rather than the regexp that was previously in use.
To accommodate this, I removed the (needless) `async` declarations of everything in `evaluator.js`.


core/expressionParser/helpers/evaluator.js
```diff
@@ -5,37 +5,98 @@ const { isString } = require('@core/stringUtils')
...
-  const leftResult = await evalExpression(left, ctx)
-  const rightResult = await evalExpression(right, ctx)
+
+  const fn = binaryOperators[operator]
+  if (!fn) throw new SystemError('undefinedFunction', { fnName: operator })
+
+  const leftResult = evalExpression(left, ctx)
+  const rightResult = evalExpression(right, ctx)
 
   if (R.isNil(leftResult) || R.isNil(rightResult))
```
eliask: One thing I wanted to ask about here: If we have e.g. a logical AND here: `null || (foo > 0)`, now it would get evaluated as null. Is this really what we want?

minotogna: Good point @eliask . I think this should be removed. The expression should be evaluated with null values as well. Let's remove the ```if (R.isNil(leftResult) || R.isNil(rightResult))``` check. what do you think?

eliask: Yeah I'll remove it.

eliask: Actually it occurred to me that we probably want something that's a bit closer to SQL semantics. So, probably no special handling for logical OR, but return null in every other case.

eliask: Hm, arguably `+` and `-` could be handled in a different way, would these semantics make sense?

```
1 + null == 1
null + 1 == 1
null - 1 == null
1 - null == 1
```

eliask: Aand there's a whole another bag of worms if we were to support text operations with `+` :)

Currently there are no type checks so it's possible for booleans, numbers, and strings to get mixed up. And we implicitly therefore support doing various operations with bool+number, number+string, etc.
Maybe no one but me would get the idea to use it, but maybe this could make sense?

```
1 + "text" == "1null" // normal JS semantics
"" + 123 == "123" // normal JS semantics
null + "" == ""
"" + null == ""
"foo" + null + "bar" == "foobar"
```

core/record/recordExpressionParser.js
```diff
@@ -8,17 +8,23 @@ const NodeDefExpression = require('@core/survey/nodeDefExpression')
...
   }
   return await Expression.evalString(query, ctx)
 }
 
+// Get reachable nodes, i.e. the children of the node's ancestors.
+// NOTE: The root node is excluded, but it _should_ be an entity, so that is fine.
+const getReachableNodes = (record, node) => {
+  const ancestors = Record.getAncestorsAndSelf(node)(record)
+  return R.flatten(ancestors.map(n => Record.getNodeChildren(n)(record)))
+}
```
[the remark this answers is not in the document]
eliask: Hm I was thinking that the data is already in memory, so it's not more than a few thousand pointers at the most?
That seems manageable to me. Maybe I'm missing something?

core/record/recordExpressionParser.js
```diff
@@ -74,7 +80,26 @@ const bindNode = (survey, record, node) => {
         : NodeDef.isBoolean(nodeDef)
           ? value === 'true'
           : value
-    }
+    },
+
+    getReachableNodeValue: nodeName => {
+      const _getNodeByGivenName = R.pipe(
```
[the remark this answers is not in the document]
eliask: I could lookup the target node by name but there's still a need to verify that the node is in the set of permissible nodes.
I don't think it would make life easier to do it that way, but I could be wrong.

core/expressionParser/helpers/evaluator.js
minotogna: For now I would leave as normal JS semantics... we'll update the behavior in the future if needed. what do you think?

core/record/recordExpressionParser.js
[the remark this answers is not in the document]
minotogna: Yes, this would work when editing  single record, but if run this for N records during survey publish it might get slow...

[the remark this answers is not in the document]
eliask: The tradeoff here is that for 10-20kB more RAM allocation for thousands of pointers, the code is (IMO) simpler.

I could benchmark this, but I suspect that the difference would be less than a millisecond vs. a version that avoid allocating an array, in the case of a survey with thousands of nodes.

Does that match what you had in mind?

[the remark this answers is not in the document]
minotogna: This can be achieved with a completely of O(1). I thinking with the hierarchy uuids with store it just simplify this. Take a look at pull request. Let us know what you think

PR #482, opened 2019-11-01
Add app version information to the web app

eliask:
The version now is basically the output of `git describe`. I tagged this commit with `0.0.0` so it displays nicely. But later versions will basically look something like `v0.0.0-34-g7c16d8b` (34 commits after 0.0.0, commit hash g7c16d8b)

So any real releases like 1.0.0 will look exactly like "1.0.0" and dev commits will also be identified with their exact commits.

See: https://www.npmjs.com/package/git-revision-webpack-plugin

The DOM element of the version string also contains additional version info in its attributes.


package.json
```diff
@@ -10,15 +10,17 @@
...
-    "dev:server": "npm run build:server:once && npm-run-all --parallel nodemon:prod watch:server",
+    "build:server:dev": "cross-env NODE_ENV=development webpack --display-error-details --config webpack.config.server.js",
+    "dev:server": "npm run build:server:dev && npm-run-all --parallel nodemon:prod watch:server",
     "watch:server": "cross-env NODE_ENV=development webpack --display-error-details --config webpack.config.server.js --watch",
     "nodemon:prod": "cross-env NODE_ENV=development node-dev --no-notify --experimental-worker --inspect -- dist/server.js",
-    "test": "mocha-webpack --timeout 30000 --display-error-details --webpack-config webpack.config.test.js \"test/*/*.js\""
+    "test": "mocha-webpack --timeout 30000 --display-error-details --webpack-config webpack.config.test.js \"test/*/*.js\"",
+    "postinstall": "webpack --display-error-details --config webpack.config.js",
+    "preversion": "yarn test",
+    "postversion": "git push --tags && yarn publish . --tag $npm_package_version && git push && echo \"Successfully released version $npm_package_version!\""
```
minotogna: I don't think we want to publish to the npm registry..

webpack.config.js
```diff
@@ -35,24 +36,31 @@ class CleanUpStatsPlugin {
...
   new MiniCssExtractPlugin({
     filename: 'styles-[hash].css'
   }),
   new HtmlWebpackPlugin({
     template: './web-resources/index.html'
   }),
   new webpack.DefinePlugin({
-    __SYSTEM_VERSION__: `"${versionString}"`,
-    __BUST__: `"${uuidv4()}"`,
+    __SYSTEM_VERSION__: JSON.stringify(versionString),
```
minotogna: I think now we can get rid of ```__SYSTEM_VERSION__```. we don't use it, we ```use process.env. APPLICATION_VERSION```

core/processUtils.js
```diff
@@ -10,12 +10,15 @@ const ENV = {
   nodeEnv: process.env['NODE_ENV'] || environments.development,
   tempFolder: process.env['TEMP_FOLDER'] || '/tmp/arena_upload',
   buildReport: process.env['BUILD_REPORT'] === 'true',
-  sourceVersion: process.env['SOURCE_VERSION'] || 'N/A',
-  //COGNITO
+  // APP VERSION
+  applicationVersion: process.env['APPLICATION_VERSION'],
+  gitCommithash: process.env['GIT_COMMIT_HASH'],
```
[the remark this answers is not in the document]
eliask: Oops, my bad, will fix

PR #494, opened 2019-11-07
Activity log analysis

fefrain:


core/i18n/resources/en.js
```diff
@@ -240,11 +240,11 @@ module.exports = {
 
       // analysis
       [ActivityLog.type.processingChainCreate]: 'added processing chain',
-      [ActivityLog.type.processingChainPropUpdate]: 'updated {{prop}} processing chain {{label}}',
+      [ActivityLog.type.processingChainPropUpdate]: 'updated {{key}} of processing chain {{label}}',
       [ActivityLog.type.processingChainDelete]: 'deleted processing chain {{label}}',
       [ActivityLog.type.processingStepCreate]: 'added processing step {{index}} to processing chain {{processingChainLabel}}',
-      [ActivityLog.type.processingStepPropsUpdate]: 'updated processing step {{index}} {{prop}} to {{value}}',
-      [ActivityLog.type.processingStepDelete]: 'deleted processing step {{index}}',
+      [ActivityLog.type.processingStepPropsUpdate]: 'updated processing step {{index}} of processing chain {{processingChainLabel}}',
```
minotogna: I would say that this message is rather: ```'updated {{key}} {{value}} of processing chain {{processingChainLabel}} step {{index}}'```
where {{key}} is entity || category and {{value}} is entity label or category name.
what do you think?

core/i18n/resources/en.js
```diff
@@ -240,11 +240,11 @@ module.exports = {
...
+      [ActivityLog.type.processingStepDelete]: 'deleted processing step {{index}} of processing chain {{processingChainLabel}}',
```
minotogna: this can be ```'deleted processing chain {{processingChainLabel}} step {{index}}'```

common/activityLog/activityLog.js
```diff
@@ -116,16 +122,19 @@ export const getDateCreated = ObjectUtils.getDateCreated
...
+export const getProcessingChainLabels = R.prop(keys.processingChainLabels)
+export const getProcessingStepIndex = R.prop(keys.processingStepIndex)
 
 // content props
 const _getContentProp = prop => R.path([keys.content, prop])
 export const getContentUuid = _getContentProp(keysContent.uuid)
 export const getContentParentUuid = _getContentProp(keysContent.parentUuid)
 export const getContentName = _getContentProp(keysContent.name)
 export const getContentKey = _getContentProp(keysContent.key)
+export const getContentValue = _getContentProp(keysContent.value)
```
minotogna: are you using ```ActivityLog. getContentValue``` method?

core/i18n/resources/en.js
fefrain: Good idea, but we pass the entire props object to the update method, we need to guess what key is being updated or modify that function and pass key and value parameters.

core/i18n/resources/en.js
fefrain: Actually I prefer the other version. Do we need to change the update message too to be consistent with it?

core/i18n/resources/en.js
minotogna: yes, entityUuid is valid when categoryUuid is null and vice versa. you just need to check which one is populated

core/i18n/resources/en.js
minotogna: which update message? this is consistent with my comment above. 
They both work, this is just shorter

core/i18n/resources/en.js
fefrain: ok

core/i18n/resources/en.js
fefrain: right

common/activityLog/activityLog.js
fefrain: the answer is inside yourself... no!

core/i18n/resources/en.js
minotogna: I implemented the merge props in the repository; in the client I set to null the one that's not used. this is still work in progress, we'll see how it will evolve

PR #495, opened 2019-11-08
CI and other improvements

eliask:


web-resources/index.html
```diff
@@ -6,6 +6,10 @@
     <meta http-equiv="Content-Type" content="text/html; charset=utf-8"/>
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <script>
+        window.cognitoUserPoolId = '$COGNITO_USER_POOL_ID';
+        window.cognitoClientId = '$COGNITO_CLIENT_ID';
+    </script>
```
minotogna: why do we need the cognito vars in window?

eliask: This is to make the built JS bundle invariant to configuration changes (Cognito parameters are currently the only such parameters), so that there's no need to build a new version when these change.

This is useful, so that e.g. a single Docker image can be used in multiple places, with different Cognito parameters. This is the closest analogue to environment variables in JS.

PR #498, opened 2019-11-08
Circular dependencies

fefrain:


webapp/commonComponents/expression/expressionVariables.js
```diff
@@ -41,37 +41,37 @@ const getSqlVariables = (nodeDef, lang) => {
...
 }
 
-const getChildDefVariables = (survey, nodeDefContext, mode, lang) => {
+const getChildDefVariables = (survey, nodeDefContext, nodeDefCurrent, mode, lang) => R.pipe(
+  Survey.getNodeDefChildren(nodeDefContext),
+  //exclude nodes that reference the current one
+  R.unless(
+    R.always(R.isNil(nodeDefCurrent)),
+    R.reject(nodeDef => Survey.isNodeDefDependentOn(NodeDef.getUuid(nodeDef), NodeDef.getUuid(nodeDefCurrent))(survey)),
+  ),
```
minotogna: the reject can be done in the map below

webapp/commonComponents/expression/expressionVariables.js
```diff
@@ -41,37 +41,37 @@ const getSqlVariables = (nodeDef, lang) => {
...
-      else if (mode === Expression.modes.json)
-        return getJsVariables(childDef, lang)
-    }),
-    R.flatten,
-    R.reject(R.isNil),
+export const getVariables = (survey, nodeDefContext, nodeDefCurrent, mode, preferredLang) => {
+  const lang = Survey.getLanguage(preferredLang)(Survey.getSurveyInfo(survey))
+  const surveyWithDependencies = R.pipe(
+    Survey.buildDependencyGraph,
+    graph => Survey.assocDependencyGraph(graph)(survey)
```
minotogna: you are building the dependency graph every time an expression is edited. is that what we really want? can this cause performance issues?

webapp/commonComponents/expression/expressionVariables.js
fefrain: Right

webapp/commonComponents/expression/expressionVariables.js
fefrain: The dependency graph can change whenever a nodeDef expression is changed; we can generate it one time and associate it to the survey in the state, but we need also to update it when needed. How can we do it?

minotogna: what do you mean how we can do it? we implement it

core/survey/_survey/surveyDependencies.js
```diff
@@ -74,7 +76,37 @@ const getNodeDefDependencies = (nodeDefUuid, dependencyType = null) => R.pipe(
...
+  while(stack.length > 0) {
+    const nodeDefUuidCurrent = stack.pop()
+
+    if (nodeDefUuid === nodeDefUuidCurrent)
+      return true
+
+    visitedUuids.add(nodeDefUuidCurrent)
+
+    const dependencies = getNodeDefDependencies(nodeDefUuidCurrent)(survey)
+    dependencies.forEach(nodeDefUuidDependent => {
```
eliask: should normal for-loops be preferred for the codebase in general?

core/survey/_surveyValidator/nodeDefValidationsValidator.js
```diff
@@ -19,11 +21,11 @@ const validate = async (survey, nodeDef, nodeDefValidations, errorKey = null) =>
   return R.pipe(
     Validation.assocFieldValidation(
       NodeDefValidations.keys.expressions,
-      await NodeDefExpressionsValidator.validate(survey, nodeDef, NodeDefValidations.getExpressions(nodeDefValidations), false)
+      await NodeDefExpressionsValidator.validate(survey, nodeDef, nodeDef, NodeDefValidations.getExpressions(nodeDefValidations), false, true)
```
eliask: I don't understand why circular references are allowed here? (the last boolean parameter)

core/survey/_survey/surveyDependencies.js
fefrain: yes, I will use Ramda forEach instead, just to be consistent with other parts of the code

core/survey/_surveyValidator/nodeDefValidationsValidator.js
fefrain: because in the node def validations you can use the current node, validating for example that it's value is greater than 0 or whatever.

PR #499, opened 2019-11-08
Sampling point data slowness fix

fefrain:


server/modules/category/manager/categoryManager.js
```diff
@@ -151,26 +151,43 @@ const updateLevelProp = async (user, surveyId, categoryUuid, levelUuid, key, val
...
       markSurveyDraft(surveyId, t),
-      ActivityLogRepository.insert(user, surveyId, ActivityLog.type.categoryItemPropUpdate, {
-        [ActivityLog.keysContent.uuid]: itemUuid,
-        [ActivityLog.keysContent.code]: CategoryItem.getCode(item),
-        [ActivityLog.keysContent.categoryUuid]: categoryUuid,
-        [ActivityLog.keysContent.levelUuid]: CategoryItem.getLevelUuid(item),
-        [ActivityLog.keysContent.key]: key,
-        [ActivityLog.keysContent.value]: value
-      }, false, t)
+      ActivityLogRepository.insertMany(user, surveyId, [_newCategoryItemUpdateLogActivity(categoryUuid, item, key, value, false)], t)
```
minotogna: why insertMany if you are inserting only one activityLog?

fefrain: It's there only to reuse the generation of the activity log with _newCategoryItemUpdateLogActivity, maybe we can have another function in ActivityLogRepository for it

minotogna: I'm not sure I understand; the method _newCategoryItemUpdateLogActivity returns one activity log. then you pass it within an array at line 172. Am I missing something?

PR #505, opened 2019-11-12
- Added WindowConfirm component

minotogna:
- Using WindowConfirm to render survey publish confirm message

### DIFF
--- a/core/i18n/resources/en.js
+++ b/core/i18n/resources/en.js
@@ -46,6 +46,7 @@ export const enTranslation = {
     no: 'No',
     noItems: `$t(common.no) $t(common.item,{'count':2})`,
     of: 'of',
+    ok: 'Ok',
     required: 'Required',
     reset: 'Reset',
     save: 'Save',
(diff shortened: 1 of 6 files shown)


core/i18n/resources/en.js
```diff
@@ -145,7 +146,18 @@ export const enTranslation = {
...
+      
+Please, note that:
+
+- All labels associated with deleted languages will be permanently deleted.
+- Node definitions cannot be changed from single to multiple and vice versa.
+- All category items cannot change code and cannot be deleted.
+- All taxa cannot be modified.
+- All records associated with deleted cycles will be permanently deleted.
+- All nodes associated with deleted node definitions will be permanently deleted.
+
```
eliask: Maybe rephrase as follows?

```
- Category items' codes cannot be changed after publishing
- Category items cannot be deleted after publishing
- Existing taxa cannot be modified after publishing
```

Although I'm not sure about the taxa. Can new taxa be added? Can they still be deleted?

Also, is "node" something that is obvious to the end user? I haven't seen "node" mentioned anywhere in the UI except for error messages.

fefrain: I agree with Elias, you should specify "after publishing" somewhere in the message, maybe in the beginning of it: Please not that after publishing: ...

webapp/commonComponents/windowConfirm.js
```diff
@@ -0,0 +1,43 @@
+import './windowConfirm.scss'
```
fefrain: I would call this module ConfirmDialog...

core/i18n/resources/en.js
minotogna: how about this

```#### You are about to publish the survey {{survey}} ####

###### During publish: ###### 
- All labels associated with deleted languages will be permanently deleted.
- All records associated with deleted cycles will be permanently deleted.
- Data associated with deleted form fields will be permanently deleted.

###### After publish: ###### 
- Form fields cannot be changed from single to multiple and vice versa.
- Category items' codes cannot be changed.
- Category items cannot be deleted.
- Taxa cannot be modified.

** Are you sure you want to continue? **```

webapp/commonComponents/windowConfirm.js
minotogna: ok

core/i18n/resources/en.js
minotogna: i changed the message

PR #515, opened 2019-11-19
Preliminary infrastructure automation with Pulumi

eliask:
This adds the CI/CD configuration that is currently being used to build the testing/QA environment at test.openforis-arena.org.

Each new tagged release should automatically get deployed to the testing environment.

Pull requests should additionally validate the Pulumi configuration to the extent that the tool can do that.

A somewhat related change is to the Travis CI configuration. Now it uses the Dockerized tests directly. This is quite a bit slower than before, but the slowness is somewhat mitigated by caching.
This change in turn is because of some difficult-to-debug problem in the difference between the Travis "native" and the Dockerized tests.


.gitignore
```diff
@@ -1,28 +1,269 @@
...
-.vscode/
-.eslintrc.json
+modules.xml
+.idea/misc.xml
+*.ipr
 
-# MacOS
-.DS_Store
+# Sonarlint plugin
+.idea/sonarlint
```
minotogna: would it make this file shorter if we would use something like:
```
.*
!/.gitignore
```

fefrain: I agree with Mino, it's easier to put the exceptions there, like .gitignore or .env.template

eliask: Hm well, there's a bunch, .eslintrc, .babelrc, .travis.yml, etc.

I just find that this approach, using curated gitignore lines from gitignore.io, reduces the effective complexity, much like using libraries in software development in general does (now there's only one manually specified line at the top).

minotogna: true, let's go for it

PR #521, opened 2019-11-20
fixed home page redirect

SteRiccio:


eliask:
Should `server/nginx.conf` also be changed in this way? Or maybe it should just be changed to point straight to the app server? I'm not sure. Otherwise OK by me.

minotogna:
> Should `server/nginx.conf` also be changed in this way? Or maybe it should just be changed to point straight to the app server? I'm not sure. Otherwise OK by me.

I think it should. probably the problem is there

fefrain:
> Should `server/nginx.conf` also be changed in this way? Or maybe it should just be changed to point straight to the app server? I'm not sure. Otherwise OK by me.

You are absolutely right...

PR #525, opened 2019-11-21
README: Added instructions to install and restart local db

minotogna:


eliask:
My bad on removing this bit. I thought it would suffice to just use docker-compose to create the DB. Though I suppose it makes it harder to use external tools with the DB. I was thinking of mainly interfacing with it like so: 
```shell
docker exec -it arena_db_1 psql --user arena of-arena-dev
```

PR #528, opened 2019-11-21
DB improvements

eliask:
- Use Bluebird to get useful stack traces for DB related errors (maybe later reconsider the options that help with debugging, when the app matures more)
- Add environment variables `DEBUG` which controls `pg-promise` statement echoing
- Make sure `postgis` exists in the DB (not the case for the AWS RDS DB by default!)
- Fix Collect import on AWS RDS (the environment contains some system triggers which couldn't be disabled, unlike in the development DB)


server/modules/collectImport/service/collectImport/dataImportJobs/recordsImportJob.js
```diff
@@ -35,7 +35,9 @@ export default class RecordsImportJob extends Job {
 
   async onStart () {
     await super.onStart()
-    await RecordManager.disableTriggers(this.surveyId, this.tx)
+
+    // Speed up processing a bit:
+    await this.tx.query('SET CONSTRAINTS ALL DEFERRED')
```
SteRiccio: I don't know if it's better to wrap this somehow with the DbUtils class...

server/modules/collectImport/service/collectImport/dataImportJobs/recordsImportJob.js
```diff
@@ -89,10 +91,6 @@ export default class RecordsImportJob extends Job {
 
   async beforeEnd () {
```
SteRiccio: You can even remove this method

server/modules/collectImport/service/collectImport/dataImportJobs/recordsImportJob.js
eliask: I wouldn't normally wrap small one-off things. I think there's such a thing as premature generalization. One rule of thumb is to start doing that once there's three or more usages of something.

PR #533, opened 2019-11-22
Prevent app crash from "expected" SystemErrors

eliask:


common/webSocket/webSocketEvents.js
```diff
@@ -13,4 +13,5 @@ export const WebSocketEvents = {
   recordDelete: 'recordDelete',
   error: 'threadError',
   recordSessionExpired: 'recordSessionExpired',
+  applicationError: 'applicationError',
```
SteRiccio: since it's an error managed only during record update, can it be something like recordUpdateError?

webapp/loggedin/surveyViews/record/actions.js
```diff
@@ -57,6 +57,11 @@ export const sessionExpired = history => dispatch => {
   _navigateToModuleDataHome(history)
 }
 
+export const applicationError = (history, recordUuid, key, params) => dispatch => {
+  dispatch(showNotification(key, params, 'error'))
```
minotogna: you can use ```NotificationState.severity.error```

common/webSocket/webSocketEvents.js
eliask: It could be, but I was wondering if there would be use for a generic non-fatal error message type

SteRiccio: sounds good

server/modules/record/service/update/thread/recordUpdateThread.js
```diff
@@ -70,7 +71,24 @@ class RecordUpdateThread extends Thread {
...
+        // SystemError is an expected error type, e.g. when there's a problem with expressions.
+        if (e instanceof SystemError) {
+          this.postMessage({
+            type: WebSocketEvents.applicationError,
+            content: {
+              key: e.key,
+              params: e.params,
+            }
+          })
+          process.exit(1) // Stop processing
```
minotogna: do you really want to process.exit ?

webapp/loggedin/surveyViews/record/actions.js
```diff
@@ -57,6 +57,11 @@ export const sessionExpired = history => dispatch => {
   _navigateToModuleDataHome(history)
 }
 
+export const applicationError = (history, recordUuid, key, params) => dispatch => {
```
minotogna: you are not using recordUuid

webapp/loggedin/surveyViews/record/recordView.js
```diff
@@ -126,7 +132,8 @@ const mapStateToProps = (state, { match, location }) => {
     recordUuidUrlParam: getUrlParam('recordUuid')(match) || recordUuidPreview,
     parentNodeUuidUrlParam: urlSearchParams.get('parentNodeUuid'),
     surveyCycleKey,
-    preview: !!recordUuidPreview
+    preview: !!recordUuidPreview,
+    applicationError,
```
minotogna: i don't think you need to pass applicationError function here...

server/modules/record/service/update/thread/recordUpdateThread.js
eliask: Right, the first time I tried to just `return`, the thread seemed to spin forever and lock the browser. But now it seems ok when I tried it again.

PR #537, opened 2019-11-26
CSVReader: reading rows iteratively, avoid recursion

minotogna:
Using CSVReader iteratively made us save approximately 80MB of memory during TaxonomiesImportJob and SamplingPointDataImportJob running within CollectImportJob


server/utils/file/csvReader.js
```diff
@@ -7,88 +7,70 @@ import * as StringUtils from '@core/stringUtils'
...
     }
 
-    const processNext = () => {
+    const _indexRowByHeaders = row => headers
+      ? headers.reduce(
+        (accRow, header, index) => Object.assign(accRow, { [header]: StringUtils.trim(row[index]) }),
+        {}
+      )
+      : row
+
```
sgiaccio: Would R.indexBy be more readable? trim would have to be done separately though

server/utils/file/csvReader.js
```diff
@@ -7,88 +7,70 @@ import * as StringUtils from '@core/stringUtils'
...
       (async () => {
-        processingRow = true
 
-        if (queue.isEmpty()) {
-          if (ended)
-            resolve()
-        } else if (!canceled) {
-          const row = queue.dequeue()
+        // run until there's a row in the queue and it's not been canceled
+        for (let row = queue.dequeue(); !!row && !canceled; row = queue.dequeue()) {
```
sgiaccio: while ((row = queue.dequeue()) !== null && !canceled) to avoid double assignment?

minotogna: ok, i changed it

server/utils/file/csvReader.js
minotogna: R.indexBy is not suitable in this context

server/utils/file/csvReader.js
```diff
@@ -7,88 +7,70 @@ import * as StringUtils from '@core/stringUtils'
...
-          if (ended)
-            resolve()
-        } else if (!canceled) {
-          const row = queue.dequeue()
+    const processQueue = () => {
+      (async () => {
 
+        let row
+        // run until there's a row in the queue and it's not been canceled
+        while (!!(row = queue.dequeue()) && !canceled) {
```
eliask: `!!` is unnecessary here, the truth-value is the same regardless.

minotogna: I know, i added it just to self document the code. If you prefer, I'll remove it

minotogna: I'll ask @SteRiccio to remove this in his next pull request

PR #538, opened 2019-11-27
Add linter and enforce code style

eliask:
Added [xo](https://github.com/xojs/xo) with configuration that more or less matches the current code style.

After the "auto-fixed" rules, there are still around 2000 linter errors or warnings left. Many of these are related to React props validation, however.

Usage: `yarn run xo`

Code style changes get auto-applied on `git commit`.


eliask:
> I think it changes too many files at once, and sometimes non it the best way (`virtual: 'virtual', //true|false` becoming `virtual: 'virtual', // True|false` for example. Would it be possible to tweak the configuration to make it a bit more permissive?

I don't know. That's the nature of automatic code formatting.
Upon first applying them there tends to be a large number of small changes, but thereafter it reaches a steady state.

I think the more important question is, what is the desired form of configuration here? In that case, perhaps the rule "Comments must begin not begin with a lower-case letter" could be disabled?

PR #545, opened 2019-12-05
Recent activities polling

SteRiccio:


webapp/loggedin/modules/home/dashboard/activityLog/activityLogView.scss
```diff
@@ -66,4 +71,13 @@
...
+}
+
+@keyframes animate-highlight {
+  from {
+    background-color: $yellow;
+  }
+  to {
+    background-color: inherit;
+  }
+}
```
minotogna: can it start inherit, 50% yellow and then go back to inherit ?

eliask:
Somehow the pre-commit hooks didn't work here since the linter rules haven't been applied. Maybe you need to run `yarn` once to install the hooks, so that commits get formatted and linted.

eliask:
OK otherwise, but the linter errors should be fixed first (that's why the CI build fails): `yarn format && yarn test` (`test` also runs the linter)

PR #548, opened 2019-12-10
Add advanced expression editor

eliask:
Add an alternative expression editor for more sophisticated expressions.

Feature a limited form of auto-completion for variable (node) names, and some function documentation for the available functions (currently just three).

The advanced editor is hidden from all but the default value expressions.


webapp/commonComponents/expression/expressionEditorPopup.js
```diff
@@ -11,15 +11,43 @@ import ExpressionNode from './nodes/expressionNode'
...
+        {...props}
+      />
+    )
+
   return (
     <Popup className="expression-editor-popup" onClose={onClose} padding={20}>
+      <button hidden={hideAdvanced} onClick={() => setAdvancedEditor(true)} style={{ position: 'absolute', right: 0 }}>
+        {i18n.t('nodeDefEdit.expressionsProp.advanced')}
+      </button>
+
```
minotogna: I would switch here the Advanced vs Basic editor. (to avoid duplicate code <Popup/>)

package.json
```diff
@@ -262,6 +263,7 @@
     "socket.io": "^2.3.0",
     "socket.io-client": "^2.3.0",
     "text-mask-addons": "^3.8.0",
+    "textarea-caret": "^3.1.0",
```
minotogna: do we still use textarea-caret?

core/i18n/resources/en.js
```diff
@@ -342,6 +342,14 @@ export const enTranslation = {
     basic: 'Basic',
     advanced: 'Advanced',
     validations: 'Validations',
+    variable: 'Variable',
+    variables: 'Variables',
```
minotogna: preferably you can use i18next convention for plural with _plural suffix. this would become:
```
variable: 'Variable',
variable_plural: 'Variables',
```

webapp/commonComponents/expression/codemirrorArenaExpressionHint.js
```diff
@@ -0,0 +1,60 @@
...
+        completions.push(getNodeCompletion(node, token))
+        break // Either node.value or node.label matches - no need for both
+      }
+    }
+  }
+
+  return completions
+}
+
+export function arenaExpressionHint(variables, editor, _options) {
```
minotogna: we usually use es6 function declarations.. what do you think?

webapp/commonComponents/expression/expressionEditorPopup.js
```diff
@@ -1,54 +1,96 @@
...
+    queryDraft,
+    exprDraft,
+    exprDraftValid,
+    updateDraft,
+    resetDraft,
+    advanced,
+    setAdvancedEditor,
+  } = useExpressionEditorPopupState({ setExpressionCanBeApplied, ...props })
+
+  const advancedEditorState = useAdvancedExpressionEditorPopupState(props)
```
minotogna: why don't you use useAdvancedExpressionEditorPopupState in the AdvancedExpressionEditorPopup directly?

eliask: I did this since currently `onApply` juggles between basic/advanced editor state and now that the base component shares the Apply/Reset buttons, it seems quite difficult to do it without access to the state at that level?

PR #550, opened 2019-12-10
Revert import ordering back to original in webapp/

eliask:


webapp/app/store.js
```diff
@@ -1,21 +1,21 @@
...
-import appReducer from './reducer'
 import notificationReducer from './appNotification/reducer'
 import errorsReducer from './appErrors/reducer'
 
+import * as LoginState from '../login/loginState'
+import * as SurveyState from '../survey/surveyState'
 import * as NotificationState from './appNotification/appNotificationState'
 import * as ErrorsState from './appErrors/appErrorsState'
 
+import createDebounce from 'redux-debounced'
```
SteRiccio: I don't understand how it works... why are these imports here?!

webapp/loggedin/modules/home/collectImportReport/components/tableRow.js
```diff
@@ -10,9 +10,10 @@ import Checkbox from '@webapp/commonComponents/form/checkbox'
 
 import * as AppState from '@webapp/app/appState'
 import * as SurveyState from '@webapp/survey/surveyState'
-import LabelsEditor from '@webapp/loggedin/surveyViews/labelsEditor/labelsEditor'
 
 import { appModuleUri, designerModules } from '@webapp/loggedin/appModules'
+import LabelsEditor from '../../../../surveyViews/labelsEditor/labelsEditor'
```
SteRiccio: It was better like it was before...!

PR #555, opened 2019-12-30
User invitation

SteRiccio:


server/modules/user/manager/userManager.js
```diff
@@ -99,7 +99,7 @@ const _updateUser = async (user, surveyId, userUuid, name, email, groupUuid, pro
 
 export const updateUser = _userFetcher(_updateUser)
 
-export const updateUsernameAndStatus = UserRepository.updateUsernameAndStatus
+export const updateNamePasswordAndStatus = UserRepository.updateNamePasswordAndStatus
```
eliask: Maybe use this?
```export {updateNamePasswordAndStatus} from '.../userRepository'```

SteRiccio: Ok, maybe we can do it even for other exports like that

PR #590, opened 2020-01-29
Processing step category editor

SteRiccio:


webapp/loggedin/modules/analysis/processingChain/components/processingChainStep.js
```diff
@@ -18,9 +20,15 @@ import ErrorBadge from '@webapp/commonComponents/errorBadge'
 const ProcessingChainStep = props => {
   const { processingStep, validation } = props
 
-  const entityUuid = ProcessingStep.getEntityUuid(processingStep)
   const survey = useSelector(SurveyState.getSurvey)
-  const entity = entityUuid ? Survey.getNodeDefByUuid(entityUuid)(survey) : null
+  const entity = R.pipe(
+    ProcessingStep.getEntityUuid,
+    R.unless(R.isNil, entityUuid => Survey.getNodeDefByUuid(entityUuid)(survey)),
```
minotogna: You might not need R.unless. Survey.getThingByUuid will be return null if the uuid you pass as parameter is null

webapp/loggedin/modules/analysis/processingChain/components/processingChainSteps.js
```diff
@@ -23,20 +24,27 @@ const ProcessingChainSteps = props => {
...
-  const processingSteps = ProcessingChain.getProcessingSteps(processingChain)
+  // Fetch steps on mount
   useEffect(() => {
     if (!ProcessingChain.isTemporary(processingChain) && !editingStep) {
       dispatch(fetchProcessingSteps(ProcessingChain.getUuid(processingChain)))
     }
   }, [])
 
+  const processingSteps = ProcessingChain.getProcessingSteps(processingChain)
+  const lastStepHasCategory = Boolean(ProcessingStep.getCategoryUuid(R.last(processingSteps)))
```
minotogna: Should we add a method to peocessingStep hasCategory?

webapp/loggedin/modules/analysis/processingChain/components/processingChainStep.js
SteRiccio: right, but it avoids a useless overhead, like getting the list of nodes object and trying to get a "null" property of it...

webapp/loggedin/modules/analysis/processingChain/components/processingChainSteps.js
SteRiccio: good

webapp/loggedin/modules/analysis/processingChain/components/processingChainStep.js
minotogna: You don't get the list of nodes.

SteRiccio: yes, you don't traverse a list but still you get a shallow copy of the nodes object and then you try to get a property "null" of it.
But since it will make the code more readable, I'll remove it.

minotogna: then the R.unless should be added to ```Survey.getThingByUuid```. what do you think?

PR #594, opened 2020-01-30
App dialog confirm

SteRiccio:


minotogna:
There's one deep scan issue to check

PR #624, opened 2020-02-21
Feature/survey creation

SteRiccio:


server/modules/survey/api/surveyApi.js
```diff
@@ -7,15 +7,20 @@ import * as JobUtils from '@server/job/jobUtils'
...
+import * as UserService from '@server/modules/user/service/userService'
 
 export const init = app => {
   // ==== CREATE
   app.post('/survey', async (req, res, next) => {
     try {
       const user = Request.getUser(req)
+      if (!User.isSystemAdmin(user)) {
+        throw new UnauthorizedError(User.getName(user))
+      }
```
minotogna: don't we have a requireAdminPermission middleware function?

SteRiccio: no but we can add it ;)

minotogna: cool

PR #628, opened 2020-03-03
Processing chain test: added ProcessingChainBuilder

SteRiccio:


test/it/analysisTest/simpleProcessingChainTest.js
```diff
@@ -2,16 +2,13 @@ import * as R from 'ramda'
...
-import * as ProcessingStep from '@common/analysis/processingStep'
-import * as ProcessingStepCalculation from '@common/analysis/processingStepCalculation'
 
-// Import * as ProcessingChainService from '@server/modules/analysis/service/processingChainService'
 import * as SurveyManager from '@server/modules/survey/manager/surveyManager'
 
 import { getContextUser } from '../../testContext'
 import * as SB from '../utils/surveyBuilder'
 import * as RB from '../utils/recordBuilder'
+import * as ChainB from '../utils/processingChainBuilder'
```
minotogna: I would call this ChainBuilder

SteRiccio: ok

test/it/analysisTest/simpleProcessingChainTest.js
```diff
@@ -8,7 +8,7 @@ import * as SurveyManager from '@server/modules/survey/manager/surveyManager'
 import { getContextUser } from '../../testContext'
 import * as SB from '../utils/surveyBuilder'
 import * as RB from '../utils/recordBuilder'
-import * as ChainB from '../utils/processingChainBuilder'
+import * as ChainB from '../utils/chainBuilder'
```
minotogna: what i meant is the import * as ChainBuilder... i don't understand a variable called ChainB. also the class itself is good chainBuilder

SteRiccio: aaa ok, it was only to make the code below shorter, avoiding to repeat ChainBuilder.chain, ChainBuilder.step etc. every time... like we did for SB and RB

PR #631, opened 2020-03-11
Processing chain aggregation: added test cases and entity aggregated view model

SteRiccio:
…view model generation


test/it/009analysisTest.js
```diff
@@ -1,5 +1,10 @@
 import * as SimpleProcessingChainTest from './analysisTest/simpleProcessingChainTest'
+import * as ComplexProcessingChainTest from './analysisTest/complexProcessingChainTest'
```
minotogna: this is just a comment:
Difficult to understand the difference between Simple and Complex. 
let's find better names later on

PR #726, opened 2020-04-19
Refactored dataTable creation using new object model

minotogna:


common/model/db/sql/index.js
```diff
@@ -1,3 +1,14 @@
+export const types = {
+  uuid: 'UUID',
+  varchar: 'VARCHAR',
+  bigint: 'BIGINT',
+  decimal: `DECIMAL(${16 + 6}, 6)`,
```
SteRiccio, line 5: is it necessary?

minotogna: we did this long time ago..

PR #776, opened 2020-05-08
Refactor analysis module (Api/Service/Manager/Repository)

minotogna:
This pull requests contains the refactor of the analysis module. The most efficient way to review it would be to check locally the branch out


server/modules/analysis/repository/chain/update.js
```diff
@@ -1,31 +1,63 @@
...
+  if (!surveyId || !chainUuid)
+    throw new Error(`Survey id and chain uuid are required. {surveyId:${surveyId}, chainUuid:${chainUuid}`)
+
   const table = new TableChain(surveyId)
 
+  const setFields = Object.keys(fields).map((field, i) =>
+    field === TableChain.columnSet.props
+      ? `${TableChain.columnSet.props} = ${TableChain.columnSet.props} || $${i + 2}::jsonb`
+      : `${field} = $${i + 2}`
+  )
```
SteRiccio, line 31: this is almost the same as the one for Calculation, can we have a generic query builder module for those queries?

server/modules/analysis/repository/step/update.js
```diff
@@ -0,0 +1,38 @@
...
+
+  const tableStep = new TableStep(surveyId)
+
+  const setFields = Object.keys(fields).map((field, i) =>
+    field === TableStep.columnSet.props
+      ? `${TableStep.columnSet.props} = ${TableStep.columnSet.props} || $${i + 2}::jsonb`
+      : `${field} = $${i + 2}`
+  )
+
+  if (setFields.length === 0) throw new Error(`At least one field is required`)
```
SteRiccio, line 30: this check should be done at the beginning on the `fields` parameter together with the other checks

server/modules/analysis/repository/chain/update.js
minotogna: we can, any suggestions where to put it and how it should be?

server/modules/analysis/repository/step/update.js
minotogna: it's the same, but I changed it

common/model/db/tables/calculation/select.js
```diff
@@ -2,15 +2,17 @@
  * Generate the select query for the processing_step_calculation table by the given parameters.
  *
  * @param {!object} params - The query parameters.
- * @param {string} [params.stepUuid=null] - The step uuid to filter.
+ * @param {string} [params.stepUuid=null] - The step uuid to by.
+ * @param {!string} [params.calculationUuid=null] - The calculation uuid to filter by.
```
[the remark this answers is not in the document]
minotogna: updated

server/modules/analysis/repository/chain/update.js
SteRiccio: it could be under db, there can be a function like buildSetPropsQuery that returns an object like {query, values} (with named parameters maybe, it would be easier to merge the parameters later).

server/modules/analysis/repository/step/update.js
SteRiccio: I know it's the same, but you should check first all the parameters, than process them.

server/modules/analysis/repository/chain/update.js
minotogna: good idea. let's do this in a separate issue

PR #807, opened 2020-05-24
DataVis: round decimal values to 2 decimal digits

### LINKED ISSUE #767: DataVis: round decimal to two digits
In data vis, when in read mode, round decimals to 2 digits

SteRiccio:
Resolves #767

### DIFF
--- a/webapp/loggedin/modules/data/dataVis/dataQuery/components/tableContent/tableColumn.js
+++ b/webapp/loggedin/modules/data/dataVis/dataQuery/components/tableContent/tableColumn.js
@@ -35,10 +35,8 @@ const TableColumn = (props) => {
 
   const getColValue = (col) => {
     const value = Object.prototype.hasOwnProperty.call(row, col) ? row[col] : null
-    if (value) {
-      if (NodeDef.isInteger(nodeDef)) return NumberUtils.formatInteger(value)
-      if (NodeDef.isDecimal(nodeDef)) return NumberUtils.formatDecimal(value)
-    }
+    if (NodeDef.isInteger(nodeDef)) return NumberUtils.formatInteger(value)
+    if (NodeDef.isDecimal(nodeDef)) return NumberUtils.formatDecimal(value)
     return value
   }
 
--- a/webapp/loggedin/modules/data/dataVis/dataQuery/components/tableContent/tableColumn.js
+++ b/webapp/loggedin/modules/data/dataVis/dataQuery/components/tableContent/tableColumn.js
@@ -2,8 +2,9 @@ import React from 'react'
 import camelize from 'camelize'
 import PropTypes from 'prop-types'
 
-import * as NodeDef from '@core/survey/nodeDef'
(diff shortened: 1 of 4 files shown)


webapp/loggedin/modules/data/dataVis/dataQuery/components/tableContent/tableColumn.js
```diff
@@ -32,6 +32,15 @@ const TableColumn = (props) => {
     return `surveyForm.${nodeDefTypePrefix}.${colName}`
   }
 
+  const getColValue = (col) => {
+    const value = Object.prototype.hasOwnProperty.call(row, col) ? row[col] : null
+    if (value && NodeDef.isDecimal(nodeDef)) {
+      // Round decimal values to 2 decimal digits
+      return Number(value).toFixed(2)
```
minotogna, line 39: should have a ```NumberUtils.round = (number, digits = 2)``` function ?

SteRiccio: ok, but maybe we can use a default value of 0 for the decimal digits because usually a round function "returns the value of a number rounded to the nearest integer". Also, in that case there can be some inconsistencies between browser when using Number.toFixed (see [this](https://stackoverflow.com/questions/566564/math-roundnum-vs-num-tofixed0-and-browser-inconsistencies))

minotogna: not sure why it should default to 0.
at this point I think it's better to format both, decimal and integer.
Let's add bignumber.js library and use space for group separators and comma for decimal. We'll set up internationalization later.
see https://mikemcl.github.io/bignumber.js/#toFor

minotogna: and we have NuberUtils.formatDecimal and formatInteger.

here the config
```
BigNumber.config({
  ERRORS: false,
  FORMAT: {
    decimalSeparator: '.',
    groupSeparator: ' ',
    groupSize: 3
  }
})
```

SteRiccio: ok good

webapp/loggedin/modules/data/dataVis/dataQuery/components/tableContent/tableColumn.js
```diff
@@ -32,6 +33,15 @@ const TableColumn = (props) => {
     return `surveyForm.${nodeDefTypePrefix}.${colName}`
   }
 
+  const getColValue = (col) => {
+    const value = Object.prototype.hasOwnProperty.call(row, col) ? row[col] : null
+    if (value) {
```
minotogna, line 38: safer to check null in NuberUtils.formatDecimal

SteRiccio: ok

PR #819, opened 2020-06-02
Test/ui test

### LINKED ISSUE #815: LoginTest
- [x] Enter wrong email / password --> check error message appear
- [x] Enter right email / password --> check user is logged in and page displayed is home dashboard

sorja:
First test case for Taiko and Gauge setup
Contains:
- Login test
- Setup for Gauge and Taiko
Note:
- install:
  - mac os: brew install gauge
  - other: https://docs.gauge.org/getting_started/installing-gauge.html?os=macos&language=javascript&ide=null
- test/browser/node_modules is symlink
Resolves #815


test/browser/tests/step_implementation.js
```diff
@@ -0,0 +1,26 @@
+/* globals beforeSuite, afterSuite */
+const { $, waitFor, openBrowser, currentURL, closeBrowser, text } = require('taiko')
+const assert = require('assert')
```
minotogna, line 3: would it be possible to use chai instead of assert?

test/browser/.gitignore
```diff
@@ -0,0 +1,8 @@
+# Gauge - metadata dir
```
minotogna, line 1: why .gitignore here?

test/browser/env/default/default.properties
```diff
@@ -0,0 +1,25 @@
+# default.properties
+# properties set here will be available to the test execution as environment variables
+
+# sample_key = sample_value
+
+#The path to the gauge reports directory. Should be either relative to the project directory or an absolute path
+gauge_reports_dir = reports
```
minotogna, line 7: can this be set as .gauge-reports ?

.eslintrc
```diff
@@ -9,6 +9,9 @@
     "mocha": true,
     "node": true
   },
+  "globals": {
+    "step": "readonly"
+  },
```
sorja, line 14: We can add globals here, helps avoid ex.
```
/* eslint-disable no-unused-expressions */
expect(await text(content).exists()).to.be.true
```

test/browser/tests/step_implementation.js
```diff
@@ -0,0 +1,41 @@
...
+      : [],
+  })
+})
+
+afterSuite(async () => {
+  await closeBrowser()
+})
+
+step('Page contains <content>', async (content) => {
+  /* eslint-disable no-unused-expressions */
```
sorja, line 29: We can add globals to .eslintrc to avoid repetition

test/browser/tests/step_implementation.js
SteRiccio: done

test/browser/env/default/default.properties
sorja: i think better to have it without dot so it's not hidden (no reason to hide)

package.json
```diff
@@ -16,6 +16,10 @@
     "create-migration": "./server/db/migration/create-migration.sh",
     "dev:server": "npm run build:server:dev && npm-run-all --parallel nodemon:prod watch:server",
     "nodemon:prod": "cross-env NODE_ENV=development node-dev --no-notify --inspect -- dist/server.js",
+    "test": "mochapack --timeout 30000 --display-error-details --webpack-config test/webpack.config.test.babel.js \"test/*/*.js\"",
+    "test:docker": "test/run_docker_tests.sh",
+    "test:browser": "gauge run -d test/browser/",
+    "preversion": "git diff --quiet || { echo \"Working directory is dirty\"; exit 1; }; yarn test:docker",
```
minotogna, line 22: preversion seems duplicate

.eslintrc
minotogna: I agree. there are many of those with expect. would it be possible to do so?

SteRiccio: the problem is not with "expect", because the result of its invocation is used by the following function invocation; the problem is with ".true" or similar. There is a plugin we can use: https://www.npmjs.com/package/eslint-plugin-chai-friendly or maybe we can think about using something different than chai (maybe assert...?)

test/browser/specs/survey.md
```diff
@@ -0,0 +1,24 @@
+# Survey specification
```
minotogna, line 1: this is survey-create test. there will be more survey tests I imagine

test/browser/tests/survey.js
```diff
@@ -0,0 +1,19 @@
+const { click, button, write, into, textBox, waitFor } = require('taiko')
+
+// TODO it should be improved: wait for the loader div to disappear
+const waitForLoadComplete = async () => {
```
minotogna, line 4: this is a common utility function. extract it to some common place

.gitignore
```diff
@@ -267,3 +267,6 @@ modules.xml
 .idea/sonarlint
 
 # End of https://www.gitignore.io/api/node,terraform,visualstudiocode,webstorm+all,macos,linux
+reports
+.gauge
+gauge-reports
```
minotogna, line 272: I don't like to idea to manually edit this generated .gitignore file. 
any better alternative?

test/browser/specs/survey.md
SteRiccio: sure, I'll reorganize them

test/browser/tests/survey.js
SteRiccio: right, I'll move it elsewhere

.gitignore
SteRiccio: it should be modified from gitignore.io, but Gauge is not among the available options.
Extra exclusions can be put on top.

minotogna: but what is reports folder?

PR #822, opened 2020-06-03
Webapp folder structure proposal

minotogna:


minotogna:
@sorja components name upper case yes, folder name it's something I would't like to see :p

SteRiccio:
> @sorja components name upper case yes, folder name it's something I would't like to see :p

I agree with Mino about the folder names, even in the documentation of React the folder names start with a lowercase letter, but it depends... if the folder is to group the files of a component and there is an index.js file inside of it, then it makes sense to have it starting with an uppercase letter to harmonize the imports.

doc/webapp.md
```diff
@@ -0,0 +1,94 @@
...
+                    - surveyForm
+                    - taxonomies
+                    - taxonomy
+    - store
+        - app
+            - dialogConfirm
+            - errors
+            - job
+            - notification
+            - i18n
```
ramirobg94, line 58: I am not sure it is the moment and the right place,

 I saw in the repo that all of the translations are into the same file, how about split it based on the view or context?

- i18n
- -  en
- - - Survey
- - - Designer
- - it
- - - ...

ramirobg94:
I think that @SteRiccio comment is a good point, I am okay in both cases but this approach works for me. Maybe this brings up what should be or not into each component folder:

components/views (as we talked this morning)
ComponentName.js
style.scss
index.js

or

components
ComponentName.js
style.scss

doc/webapp.md
minotogna: yes I know, it's already in the refactor work plan. here only webapp code refactor

PR #823, opened 2020-06-03
Feature/show ownername in my surveys

### LINKED ISSUE #809: "MY SURVEYS" - name of creator
It would be useful to see in the survey list that who has created the survey

ramirobg94:
Resolves #809

### DIFF
--- a/core/objectUtils.js
+++ b/core/objectUtils.js
@@ -11,7 +11,6 @@ export const keys = {
   id: 'id',
   index: 'index',
   name: 'name',
-  ownerName: 'ownerName',
   nodeDefUuid: 'nodeDefUuid',
   parentUuid: 'parentUuid',
   props: 'props',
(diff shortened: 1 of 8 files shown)


webapp/loggedin/modules/home/surveyList/components/surveyListRowHeader.js
```diff
@@ -13,6 +13,7 @@ const SurveyListRowHeader = () => {
       <div>{i18n.t('common.dateCreated')}</div>
       <div>{i18n.t('common.dateLastModified')}</div>
       <div>{i18n.t('homeView.surveyList.status')}</div>
+      <div>{i18n.t('common.owner')}</div>
```
minotogna, line 16: Column should be placed after survey name

core/objectUtils.js
```diff
@@ -31,6 +32,7 @@ export const isEqual = other => self => getUuid(other) === getUuid(self)
 // ====== READ
 export const getId = R.prop(keys.id)
 export const getUuid = R.propOr(null, keys.uuid)
+export const getOwnerName = R.propOr(null, keys.ownerName)
```
minotogna, line 35: I don't see getOwnerName as common prop. Add it only to SurveyInfo

ramirobg94: sure!

PR #824, opened 2020-06-03
replace logger with dev tools

ramirobg94:
As we talked this morning I did this PR to think if we like redux-dev-tools or redux-logger


package.json
```diff
@@ -64,6 +64,7 @@
     "npm-run-all": "^4.1.5",
     "optimize-css-assets-webpack-plugin": "^5.0.1",
     "prettier": "^2.0.4",
+    "redux-devtools-extension": "^2.13.8",
     "redux-logger": "^3.0.6",
```
SteRiccio, line 68: will redux-logger still be needed?

webapp/app/store.js
```diff
@@ -38,13 +39,13 @@ const createReducer = (asyncReducers) =>
...
-}
-
 // App store
-export const store = createStore(createReducer({}), applyMiddleware(...middleware))
+export const store = createStore(
+    createReducer({}),
+    composeWithDevTools(
+      applyMiddleware(...middleware)
+    )
+)
```
sorja, line 48: It is a good idea to check if the browser has the extension, to avoid getting undefined error:
`typeof window === 'object' && window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__`
https://github.com/zalmoxisus/redux-devtools-extension/issues/320

webapp/app/store.js
```diff
@@ -20,6 +20,7 @@ import * as AppErrorsState from '@webapp/app/appErrors/appErrorsState'
 import * as LoginState from '@webapp/guest/login/loginState'
 import * as SurveyState from '@webapp/survey/surveyState'
 
+
 const appReducers = {
```
sorja, line 24: Maybe one empty line is enough here?

package.json
sorja: If we want, we can use redux-logger as fallback in case the browser doesn't have the extension?
I'd prefer not though

ramirobg94: me neither,

should we remove the dependency?

webapp/app/store.js
ramirobg94: hehe :)

package.json
minotogna: yes remove the dependency from package.json

PR #835, opened 2020-06-04
DataVis: fixed error fetching data from virtual entity (uuid column not found)

### LINKED ISSUE #834: DataVis: error fetching data from virtual entity (column virtualEntityName_uuid does not exist)

SteRiccio:
Fixed #834


common/model/db/views/dataNodeDef/view.js
```diff
@@ -44,12 +44,14 @@ export default class ViewDataNodeDef extends TableDataNodeDef {
...
+  get columnNodeDefUuid() {
+    // node def can be the entity itself or it's source entity (if virtual)
+    const { nodeDef } = this.tableData
+    return new ColumnNodeDef(this, nodeDef)
   }
 
-  get columnNodeDefUuid() {
-    return new ColumnNodeDef(this, this.nodeDef)
+  get columnUuid() {
+    return this.columnNodeDefUuid.namesFull[0]
```
minotogna, line 54: I would add a method to column nameFull. what do u think?

SteRiccio: ok

PR #844, opened 2020-06-05
Fixed error publishing survey when root entity key attribute has been changed

### LINKED ISSUE #837: NodeDefEdit: Root entity key update
Removing the root entity key has several issues listed below

**Context**
- A survey with two cycles where the root entity key has name cluster_id and type is text
- Survey has been published and has already several records.

**Steps**
- Delete cluster_id 
- Add another key attribute  cluster_id 
  - type: code 
  - cycles: 1, 2
  - display as: dropdown

**Issues**
- [ ] Select cycle 2 --> Click edit cluster_id --> Display as has no value. It must be dropdown
- [x] Click publish survey --> Record Check job gives error: TypeError: Cannot read property 'deleted' of undefined

SteRiccio:
Quick and dirty solution that fixes #837.
webapp/survey/nodeDefs/actions.js file needs to be splitted asap into separate files for different action types (create/update/delete).


core/survey/_survey/surveyNodeDefs.js
```diff
@@ -81,7 +81,7 @@ export const getNodeDefSiblingByName = (nodeDef, name) => (survey) => {
 export const getNodeDefKeys = (nodeDef) =>
   R.pipe(
     getNodeDefChildren(nodeDef),
-    R.filter((n) => NodeDef.isKey(n))
+    R.filter((n) => NodeDef.isKey(n) && !NodeDef.isDeleted(n))
```
minotogna, line 84: now that I think about, should the ```!NodeDef.isDeleted(n)``` check be added to ```getNodeDefChildren``` instead?

PR #848, opened 2020-06-07
Refactor/survey store

minotogna:
This refactor is about moving the survey store to store/survey and import/export of State,Actions,Reducer
Functionalities haven't changed.

Important changes are:

- **Moved Arena to views**
  - [Main.js](https://github.com/openforis/arena/pull/848/files#diff-281d9674751de7bd55671af630487ca3)
  - [Arena.js](https://github.com/openforis/arena/pull/848/files#diff-914f31bea16166f0ee4189c52d54aedb)
- **How to export combined State/Actions** from combined reducer
  - [Survey store](https://github.com/openforis/arena/pull/848/files#diff-a1961d5e236584c99fafdcf70ed31235)

All the other changes are about State/Actions import and usage
e.g. 
```
import { SurveyState, NodeDefsActions } from '@webapp/store/survey'
...
await dispatch({ type: NodeDefsActions.nodeDefCreate, nodeDef })
```


webapp/loggedin/modules/analysis/calculation/actions/state.js
```diff
@@ -5,11 +5,10 @@ import * as NodeDef from '@core/survey/nodeDef'
 import * as Step from '@common/analysis/processingStep'
 import * as Calculation from '@common/analysis/processingStepCalculation'
 
-import * as SurveyState from '@webapp/survey/surveyState'
+import { SurveyState, NodeDefsActions } from '@webapp/store/survey'
```
ramirobg94, line 8: It looks weird to me?
What is the relation between Survey and NodeDef?

minotogna: NodeDefs, Categories, Taxonomies are all survey metadata . Accessed through Survey . We can talk about it later

webapp/store/survey/actions/actionTypes.js
```diff
@@ -0,0 +1,5 @@
+export const surveyCreate = 'survey/create'
+export const surveyUpdate = 'survey/update'
+export const surveyDelete = 'survey/delete'
+export const surveyDefsLoad = 'survey/defs/load'
+export const surveyDefsReset = 'survey/defs/reset'
```
ramirobg94, line 5: UPPER or lower case this constants?

webapp/store/survey/actions/active.js
```diff
@@ -0,0 +1,12 @@
+import axios from 'axios'
+
+import { surveyCreate, surveyUpdate } from './actionTypes'
+
+export const setActiveSurvey = (surveyId, canEdit = true, dispatchSurveyCreate = false) => async (dispatch) => {
```
ramirobg94, line 5: How about use the params as objects to avoid problems with the order?
({ surveyId, canEdit, ...})

webapp/store/survey/actions/active.js
```diff
@@ -0,0 +1,12 @@
...
+  // Load survey
+  const params = { draft: canEdit, validate: canEdit }
+  const {
+    data: { survey },
+  } = await axios.get(`/api/survey/${surveyId}`, { params })
```
ramirobg94, line 10: Maybe it is not the right moment but we have axios spread around the code how about make (a function/set of functions) as a proxy of axios? I mean like `api.get` and this Api object wrapper a call to axios, in that way it is easier to replace axios.

webapp/store/survey/actions/actionTypes.js
minotogna: not now

webapp/store/survey/actions/active.js
minotogna: yes but not now

PR #855, opened 2020-06-08
status

ramirobg94:


webapp/app/actions.js
```diff
@@ -18,7 +17,7 @@ export const initApp = () => async (dispatch) => {
   const { user, survey } = await _fetchUserAndSurvey()
   dispatch({
     type: appPropsChange,
-    status: AppState.appStatus.ready,
+    status: SystemStatusState.systemStatus.ready,
```
ramirobg94, line 20: systemStatus or systemStatuses?

minotogna: systemStatus it's ok

PR #860, opened 2020-06-09
Refactor/user

ramirobg94:
There are still some actions that we need to move. 
- initApp

The old app state was deleted with its reducer

useAuth, useUser and useProfilePicture were moved into user hooks

**we need to move this**
`_fetchUserAndSurvey`


webapp/store/user/actions.js
```diff
@@ -0,0 +1,13 @@
+import axios from 'axios'
+import { LoaderActions } from '@webapp/store/ui'
+
+export const APP_USER_LOGOUT = 'app/user/logout'
```
minotogna, line 4: should this action be just: ```export const USER_LOGOUT = 'user/logout'```

webapp/app/actions.js
```diff
@@ -1,13 +1,17 @@
...
-export const appUserLogout = 'app/user/logout'
+
+const _fetchUserAndSurvey = async () => {
+  const {
+    data: { user, survey },
+  } = await axios.get('/auth/user')
+  return { user, survey }
+}
 
 // ====== INIT
```
ramirobg94, line 17: this is going to be refactored to `initSystem` dispatching an action `SYSTEM_INIT = 'system/init'`

webapp/app/actions.js
```diff
@@ -1,13 +1,17 @@
...
 
 export const appPropsChange = 'app/props/change'
-export const appUserLogout = 'app/user/logout'
+
+const _fetchUserAndSurvey = async () => {
+  const {
+    data: { user, survey },
+  } = await axios.get('/auth/user')
+  return { user, survey }
+}
```
ramirobg94, line 14: we need to find a place for this but maybe the `initSystem` should dispatch `initUser` which fetch survey and user too

minotogna: ok in this case yes

PR #863, opened 2020-06-09
initSytem

ramirobg94:


webapp/store/survey/surveyInfo/reducer.js
```diff
@@ -3,7 +3,7 @@ import { exportReducer } from '@webapp/utils/reduxUtils'
 import * as Survey from '@core/survey/survey'
 
 // App actions
-import { appPropsChange } from '@webapp/app/actions'
+import * as SystemActions from '@webapp/store/system/actions'
```
minotogna, line 6: this must be
```import { SystemActions } from '@webapp/store/system'```

File: webapp/store/system/status/reducer.js [hunk 2]
```diff
@@ -1,9 +1,10 @@
 import { exportReducer } from '@webapp/utils/reduxUtils'
 
-import { appPropsChange } from '@webapp/app/actions'
+import * as SystemActions from '@webapp/store/system/actions'
```
minotogna, line 3: this must be local import

webapp/store/user/reducer.js
```diff
@@ -1,13 +1,13 @@
 import { exportReducer } from '@webapp/utils/reduxUtils'
 
-import { appPropsChange } from '@webapp/app/actions'
+import * as SystemActions from '@webapp/store/system/actions'
 import * as SurveyActions from '@webapp/store/survey/actions'
```
minotogna, line 4: they should be both imported from index

PR #870, opened 2020-06-11
Refactor/table component

minotogna:
Refactored Table component and moved it under webapp/components

**[- Main interesting change is the local state usage instead of redux](https://github.com/openforis/arena/pull/870/files#diff-160fd7ecb07989945c323360be0d4061)**


webapp/components/Table/Content/Content.js
```diff
@@ -44,7 +42,7 @@ const TableContent = (props) => {
               className={className}
               style={{ gridTemplateColumns }}
             >
-              {React.createElement(rowComponent, { ...props, idx: i, offset, row })}
+              {React.createElement(rowComponent, { idx: i, offset, row, rowNo: i + offset + 1, active })}
```
ramirobg94, line 45: should we propagate the rest of props as we did before?

webapp/components/Table/Header/Paginator/Paginator.js
```diff
@@ -0,0 +1,54 @@
...
+const Paginator = (props) => {
+  const { offset, limit, count } = props
+  const currentPage = offset / limit + 1
+  const totalPage = Math.ceil(count / limit)
+
+  const i18n = useI18n()
+
+  return (
+    <div className="table__paginator">
+      <Link className="btn btn-transparent" aria-disabled={count < limit || currentPage === 1} to={getLink(0)}>
```
ramirobg94, line 17: `currentPage === 1`

should we extract this to `isFirstPage`?

webapp/components/Table/tableLink.js
```diff
@@ -0,0 +1,12 @@
+export const getLimit = () => 15
```
ramirobg94, line 1: should we store this limit as constant?

minotogna: nope. this will be needed. I will explain later why this will be dynamic

webapp/components/Table/Header/Paginator/Paginator.js
minotogna: ok done

webapp/components/Table/Content/Content.js
minotogna: nope

PR #876, opened 2020-06-12
input should be uncontrolled

ramirobg94:


ramirobg94:
Fixed but could be nice if we can check if backend still works in that part

webapp/components/form/input.js
```diff
@@ -30,7 +30,7 @@ const InputComponent = (props, ref) => {
           mask={mask}
           className="form-input"
           aria-disabled={disabled}
-          isControlled={true}
+          isControlled={false}
           value={value}
           onChange={(e, { value }) => onChangeValue(value)}
```
sorja, line 35: Should the naming convention be in these cases `event`?
Although, now when not used maybe replace it with `_`

ramirobg94:
The solution proposed has some negative side effects which make us lose functionalities

PR #898, opened 2020-06-21
Refactor/chain/create

ramirobg94:
This PR have more branches:

Chain:
- [x] New chain
- [x] Update chain props
- [x] Update cycles
- [x] Edit old branch
- [x] Cancel/dismiss
- [x] Delete

Steps:
- List
    - [x] New step
    - [x] Select Step
StepItem
    - [x] Update
    - [x] New step
    - [x] Add virtual entity/edit
    - [x] CategorySelector
    - [x] EntitySelector
    - [x] Delete/cancel

Calculations
- List
    - [x] New calculations
    - [x] Drag calculations
    - [x] Select
 - Calculation
    - [x] New
    - [x] Delete
    - [x] UpdateProps
    - [x] UpdateAttbts
    - [x] Create/edit nodedef analysis

RStudioButton

Buttons
- [x] Save
- [x] Delete
- [x] Cancel/back

Styles,
Clean hooks/Storage


webapp/service/storage/index.js
```diff
@@ -1 +1,2 @@
 export { SidebarActions, useIsSidebarOpened } from './sidebar'
+export { AnalysisActions } from './analysis'
```
minotogna, line 2: I'm thinking that both should be exported only from their own folder.
what do u think?

webapp/views/App/views/Analysis/Chains/HeaderLeft/HeaderLeft.js
```diff
@@ -1,17 +1,17 @@
...
 const HeaderLeft = () => {
-  const dispatch = useDispatch()
   const history = useHistory()
   const i18n = useI18n()
 
+  const onClickNew = () => history.push(`${appModuleUri(analysisModules.processingChain)}`)
+
   return (
-    <button type="button" onClick={() => dispatch(createChain(history))} className="btn btn-s">
+    <button type="button" onClick={onClickNew} className="btn btn-s">
```
minotogna, line 14: you can use ```<Link to={appModuleUri(analysisModules.processingChain)} className='btn btn-s' />```

webapp/views/App/views/Analysis/Chain/Chain.scss
```diff
@@ -0,0 +1,4 @@
+.chain {
+
```
minotogna, line 2: style is missing

webapp/views/App/views/Analysis/Chain/Chain.js
```diff
@@ -0,0 +1,59 @@
...
+import { useI18n } from '@webapp/store/system'
+
+import { useChain } from './store'
+
+const ChainComponent = () => {
+  const i18n = useI18n()
+  const { chain, dirty, onUpdate, onSave, onDismiss } = useChain()
+  const validation = Chain.getValidation(chain)
+
+  return (
```
minotogna, line 21: this component is not like the current one...
ButtonRStudio is missing and I would extract the ButtonBar as it was before

webapp/service/storage/index.js
ramirobg94: I am thinking that maybe it is better do 3 modules inside the analysis folder one for each entity, are you talking about that? Or just put the three entities at the first level?

ramirobg94: Also in the next pr which is the pr of the steps I have the feeling that I need to extract the actions in that way too

webapp/views/App/views/Analysis/Chains/HeaderLeft/HeaderLeft.js
ramirobg94: Yes! Now we don’t need the action because we don’t do anything else

webapp/views/App/views/Analysis/Chain/Chain.scss
ramirobg94: I have to refactor all the styles because now they are imported into the same file, but I think for now I will remove this file. Because it is not needed

webapp/views/App/views/Analysis/Chain/Chain.js
ramirobg94: I did both things into the next branch, should I do into this too?

webapp/service/storage/index.js
minotogna: no I meant when you import them, you import from ```import {AnalysisActions} from '@webapp/service/storage/analysis' not from storage

PR #904, opened 2020-06-24
Refactor/data explorer

minotogna:
Data Explorer Pull request #2

- Removed Data redux store
- Moved DataQuery to common components 
- Added Query model object


common/model/query/query.js
```diff
@@ -0,0 +1,47 @@
...
+// mode
+const isMode = (mode) => (query) => getMode(query) === mode
+export const isModeAggregate = isMode(modes.aggregate)
+export const isModeRawEdit = isMode(modes.rawEdit)
+
+// utils
+export const hasSelection = (query) =>
+  isModeAggregate(query)
+    ? !A.isEmpty(getEntityDefUuid(query)) && !A.isEmpty(getMeasures(query)) && !A.isEmpty(getDimensions(query))
+    : !A.isEmpty(getEntityDefUuid(query)) && !A.isEmpty(getAttributeDefUuids(query))
```
ramirobg94, line 32: could be?

```suggestion
export const hasSelection = (query) =>
  !A.isEmpty(getEntityDefUuid(query)) &&
  (isModeAggregate(query)
    ? !A.isEmpty(getMeasures(query)) && !A.isEmpty(getDimensions(query))
    : !A.isEmpty(getAttributeDefUuids(query)))
```

common/model/query/query.js
```diff
@@ -0,0 +1,47 @@
...
+
+// mode
+export const toggleModeAggregate = (query) => ({
+  ...create({ entityDefUuid: getEntityDefUuid(query) }),
+  [keys.mode]: isModeAggregate(query) ? modes.raw : modes.aggregate,
+})
+export const toggleModeEdit = (query) => ({
+  ...query,
+  [keys.mode]: isModeRawEdit(query) ? modes.raw : modes.rawEdit,
+})
```
ramirobg94, line 47: why an entityDef is created with you toggleModeAggregate?

should if be created when you switch from aggregate to raw?

webapp/components/DataQuery/ButtonBar/ButtonBar.js
```diff
@@ -0,0 +1,73 @@
...
+          className={classNames('btn', 'btn-s', 'btn-edit', { highlight: Query.isModeAggregate(query) })}
+          onClick={() => onChangeQuery(Query.toggleModeAggregate(query))}
+          aria-disabled={appSaving || modeEdit || !nodeDefsSelectorVisible}
+        >
+          <span className="icon icon-sigma icon-14px" />
+        </button>
+        {canEdit && hasSelection && (
+          <button
+            type="button"
+            className={`btn btn-s btn-edit${modeEdit ? ' highlight' : ''}`}
```
ramirobg94, line 46: shall we use this **classNames**?
I personally don't like it because it is not plain html/jsx but it is like when we extract the className calculation.

webapp/components/DataQuery/ButtonBar/ButtonBar.js
```diff
@@ -0,0 +1,73 @@
...
+            onClick={() => onChangeQuery(Query.toggleModeEdit(query))}
```
ramirobg94, line 47: should be toggleModeEdit and the others actions per se, what happens if we click in both places "at the same time":

onClick={() => onChangeQuery(Query.Fun_A(query))}
onClick={() => onChangeQuery(Query.Fun_B(query))}

minotogna: how can a user click at the same time two buttons?

webapp/components/DataQuery/ButtonBar/ButtonBar.js
minotogna: I will use classNames in next PR. why don't you like it? makes the code simpler

common/model/query/query.js
minotogna: when switching from raw to aggregate mode and viceversa, only entity selected remains in query object. all the rest is reset

common/model/query/query.js
minotogna: same but yes I will update in next PR

PR #908, opened 2020-06-25
DataQuery: added fetchData action using rxjs

minotogna:
Added data fetch to DataQuery using RxJS

### DIFF
--- a/webapp/components/hooks/usePost.js
+++ b/webapp/components/hooks/usePost.js
@@ -1,31 +1,37 @@
 import { useEffect, useState } from 'react'
 import { useDispatch } from 'react-redux'
 
-import { asyncScheduler, Subject } from 'rxjs'
-import { switchMap, throttleTime } from 'rxjs/operators'
+import { asyncScheduler, BehaviorSubject, EMPTY, Subject } from 'rxjs'
+import { map, switchMap, throttleTime } from 'rxjs/operators'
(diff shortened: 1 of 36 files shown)


webapp/components/hooks/usePost.js
```diff
@@ -0,0 +1,37 @@
...
+
+import { asyncScheduler, Subject } from 'rxjs'
+import { switchMap, throttleTime } from 'rxjs/operators'
+import { ajax } from 'rxjs/ajax'
+import { fromAjax } from '@webapp/utils/rxjs'
+
+const createAjaxSubject = ({ dispatch, throttle }) => {
+  const fetchSubject = new Subject()
+  return fetchSubject.pipe(
+    throttleTime(throttle, asyncScheduler, { leading: false, trailing: true }),
```
ramirobg94, line 12: what is exactly the meaning of leading and trailing in this context.

I read [that](https://rxjs-dev.firebaseapp.com/api/operators/throttleTime) and the configuration looks like the double click example which listens to the second click of every tuple of click which happens between 400ms but I think that I am not able to understand it yet

minotogna: it takes the trailing event, not the leading one.
e.g. [e1,e2,e3] with leading true, e1 goes through, with trailing true, e3 goes through

PR #911, opened 2020-06-25
NodeDefDetails: added useSetNodeDefProp hook

SteRiccio:


webapp/components/survey/NodeDefDetails/store/actions/types.js
```diff
@@ -0,0 +1,3 @@
+export default {
+  NODE_DEF_PROPS_UPDATE: 'survey/nodeDef/props/update',
```
minotogna, line 2: are u using useReducer?

SteRiccio: right, I'll update it

PR #914, opened 2020-06-26
DataQuery: added table paginator

minotogna:


webapp/components/DataQuery/Visualizer/Visualizer.js
```diff
@@ -1,37 +1,26 @@
...
-          offset,
-        })
-      )}
+      {React.createElement(components[Query.getDisplayType(query)], {
+        query,
+        data,
+        dataEmpty,
+        nodeDefsSelectorVisible,
+        offset,
+      })}
```
ramirobg94, line 23: are these values going to be the same in each kind of component?

minotogna: at the moment we have only one (Table) and that's what it needs.
But I guess yes, when we'll add other visualization components they won't need more than that

webapp/components/DataQuery/store/actions/useFetchCount.js
```diff
@@ -0,0 +1,30 @@
...
+  return {
+    fetchCount: useCallback(
+      ({ query }) =>
+        post({
+          url: `${getUrl({ surveyId, query })}/count`,
+          body: getBody({ cycle, query }),
+        }),
+      [cycle, surveyId, post]
+    ),
+    resetCount: useCallback(() => reset(), [reset]),
```
SteRiccio, line 28: can it be just `resetCount: useCallback(reset, [reset])`?

minotogna: it can. will update in a next branch

PR #921, opened 2020-06-26
Refactor/chain/detele

ramirobg94:


webapp/store/survey/actions/actionTypes.js
```diff
@@ -3,3 +3,5 @@ export const surveyUpdate = 'survey/update'
 export const surveyDelete = 'survey/delete'
 export const surveyDefsLoad = 'survey/defs/load'
 export const surveyDefsReset = 'survey/defs/reset'
+export const surveyChainSave = 'survey/chain/save'
+export const surveyChainElementDelete = 'survey/chain/delete'
```
minotogna, line 7: this is better: ```export const surveyChainItemDelete = 'survey/chain/item/delete'```

webapp/store/survey/actions/chainElementDelete.js
```diff
@@ -0,0 +1,3 @@
+import * as ActionTypes from './actionTypes'
+
+export const chainElementDelete = () => ({ type: ActionTypes.surveyChainElementDelete })
```
minotogna, line 3: ```export const chainItemDelete = () => ({ type: ActionTypes.surveyChainElementDelete })```

and add it to webapp/store/survey/actions/chain.js

webapp/store/survey/actions/chainSave.js
```diff
@@ -0,0 +1,3 @@
+import * as ActionTypes from './actionTypes'
+
+export const chainSave = () => ({ type: ActionTypes.surveyChainSave })
```
minotogna, line 3: add this to webapp/store/survey/actions/chain.js

webapp/store/survey/actions/actionTypes.js
ramirobg94: Thanks, I had some bad feelings when I put that name

PR #923, opened 2020-06-28
DataExplorer: Enabled data sorting 

minotogna:
Added panel to sort data in Explorer

Screenshot 2020-06-28 at 12 54 04


webapp/components/form/dropdown.js
```diff
@@ -63,25 +63,16 @@ class Dropdown extends React.Component {
...
-    const { items } = this.state
+    const { onBeforeChange, onChange, clearOnSelection } = this.props
 
-    let changed = false
     if (!onBeforeChange || (await onBeforeChange(item))) {
+      this.setState((statePrev) => ({
+        ...statePrev,
+        displayValue: clearOnSelection ? '' : this.getItemLabel(item),
+        opened: false,
+      }))
```
SteRiccio, line 73: I don't know if prevState is necessary here, it can be just
`this.setState({
  displayValue: clearOnSelection ? '' : this.getItemLabel(item),
  opened: false
})`

minotogna: that specific fix drove me crazy. The component was setting the state after on change; that means it was overriding props passed from outside after onChange.
it must be refactored anyway

common/model/query/query/defaults.js
```diff
@@ -1,10 +1,12 @@
+import { Sort } from '../sort'
+
 import { displayTypes, keys, modes } from './keys'
 
 export const defaults = {
   [keys.mode]: modes.raw,
   [keys.displayType]: displayTypes.table,
   [keys.filter]: null,
-  [keys.sort]: [],
+  [keys.sort]: Sort.create(),
```
ramirobg94, line 9: what is the motivation to use a function?

common/model/query/sort/sort.js
```diff
@@ -0,0 +1,34 @@
+import * as A from '@core/arena'
+
+import { SortCriteria } from '../sortCriteria'
+
+export const create = () => []
+
+// ====== READ
+export const { isEmpty } = A
+export const containsVariable = (variable) => (sort) =>
+  Boolean(sort.find((sortCriteria) => SortCriteria.getVariable(sortCriteria) === variable))
```
ramirobg94, line 10: is find already boolean?

minotogna: yes damn it sorry... will change in the next PR

common/model/query/query/defaults.js
minotogna: to do not expose the model state

PR #975, opened 2020-07-07
Categories Refactor

### LINKED ISSUE #970: Categories Refactor
- [x] Update Api endpoint to match Table format
- [ ] Refactor UI using Table component (Do not use Items)

SteRiccio:
Resolves #970 
It combines 2 pull requests already approved:
- [x] Update API endpoint to match Table format
- [x] Refactor UI using Table component (Do not use Items)


server/modules/category/repository/categoryRepository.js
```diff
@@ -1,11 +1,12 @@
...
+import * as DB from '../../../db'
+import { db } from '../../../db/db'
+import * as DbUtils from '../../../db/dbUtils'
 
-import * as Category from '@core/survey/category'
-import * as CategoryLevel from '@core/survey/categoryLevel'
-import * as CategoryItem from '@core/survey/categoryItem'
+import * as Category from '../../../../core/survey/category'
+import * as CategoryLevel from '../../../../core/survey/categoryLevel'
+import * as CategoryItem from '../../../../core/survey/categoryItem'
```
ramirobg94, line 10: are there any reason to make the path relative?

server/modules/category/api/categoryApi.js
```diff
@@ -96,23 +96,61 @@ export const init = app => {
...
   })
 
+  app.get('/survey/:surveyId/categories', AuthMiddleware.requireSurveyViewPermission, async (req, res, next) => {
+    try {
+      const { surveyId, draft, validate, offset = 0, limit = null } = Request.getParams(req)
+
+      const list = await CategoryService.fetchCategoriesBySurveyId({
+        surveyId,
+        draft,
+        includeValidation: validate,
```
minotogna, line 127: maybe the param of CategoryService.fetchCategoriesBySurveyId can also be called validate

server/modules/category/repository/categoryRepository.js
SteRiccio: It prepares the process of getting rid of the server side compilation...

server/modules/category/api/categoryApi.js
SteRiccio: I prefer to keep it as it is, it does not actually validate but only fetch the categories with validation. Maybe the parameter name can be switched between the service layer and the manager.

PR #996, opened 2020-07-15
Add common button

ramirobg94:
resolves #996 
Not sure if it is better this approach or the previous one, 
Before it was the responsibility of the init and if at some moment we like to move to avoid the creation until the user saves is going to be easier.


minotogna:
we could always set a temp flag to the db until the user presses save (faster implementation if we want to use the save button login)

webapp/components/survey/ButtonTaxonomyAdd/ButtonTaxonomyAdd.js
```diff
@@ -0,0 +1,57 @@
...
+    } = await axios.post(`/api/survey/${surveyId}/taxonomies`, Taxonomy.newTaxonomy())
+
+    dispatch(SurveyActions.metaUpdated())
+
+    if (inTaxonomiesPath || A.isNull(onTaxonomyCreated)) {
+      history.push(`${appModuleUri(designerModules.taxonomy)}${Taxonomy.getUuid(taxonomyCreated)}`)
+    } else {
+      onTaxonomyCreated(taxonomyCreated)
+    }
+  }, [])
```
minotogna, line 39: this is responsibility of the caller. pass an handler `onAdd`

webapp/components/survey/ButtonTaxonomyAdd/ButtonTaxonomyAdd.js
```diff
@@ -0,0 +1,44 @@
...
+
+  const dispatch = useDispatch()
+
+  const { onAdd } = props
+
+  const surveyId = useSurveyId()
+
+  const add = useCallback(async () => {
+    const {
+      data: { taxonomy: taxonomyCreated },
```
minotogna, line 24: you could have left taxonomy. no need to override the name..

webapp/components/survey/ButtonTaxonomyAdd/ButtonTaxonomyAdd.js
```diff
@@ -0,0 +1,42 @@
...
+import React, { useCallback } from 'react'
+import { useDispatch } from 'react-redux'
+import PropTypes from 'prop-types'
+
+import * as Taxonomy from '@core/survey/taxonomy'
+
+import { useI18n } from '@webapp/store/system'
+import { SurveyActions, useSurveyId } from '@webapp/store/survey'
+
+const ButtonTaxonomyAdd = (props) => {
```
minotogna, line 12: make it generic. ButtonMetaItemAdd

webapp/components/survey/TaxonomyDetails/store/actions/useInit.js
```diff
@@ -1,43 +1,22 @@
...
-        taxonomyToSet = taxonomyCreated
-        dispatch(SurveyActions.metaUpdated())
-      } else {
-        const {
-          data: { taxonomy: taxonomyFetched },
-        } = await axios.get(`/api/survey/${surveyId}/taxonomies/${taxonomyUuidParam}?draft=true&validate=true`)
-        taxonomyToSet = taxonomyFetched
-      }
+  return useCallback(async ({ taxonomyUuid }) => {
+    if (!A.isEmpty(taxonomyUuid) || !A.isEmpty(taxonomyUuidParam)) {
```
minotogna, line 17: I would still pass only one taxonomyUuid from the component.
And it cannot be empty.
What do you think?

ramirobg94: solved

webapp/components/survey/TaxonomyDetails/store/actions/useInit.js
```diff
@@ -1,43 +1,16 @@
...
-        const {
-          data: { taxonomy: taxonomyFetched },
-        } = await axios.get(`/api/survey/${surveyId}/taxonomies/${taxonomyUuidParam}?draft=true&validate=true`)
-        taxonomyToSet = taxonomyFetched
-      }
-    }
-
-    setState(State.create({ taxonomy: taxonomyToSet }))
+  return useCallback(async ({ taxonomyUuid }) => {
+    const taxonomy = await API.fetchTaxonomy({ surveyId, Uuid: taxonomyUuid })
```
minotogna, line 13: why uuid upper case?

PR #1006, opened 2020-07-23
Bugfix - Category details validation

### LINKED ISSUE #842: CategoryEdit: validation
- [x] CategoryEdit: code item errorBadge doesn't have icon
Screenshot 2020-06-05 at 16 06 22

- [x] Survey publish: error label is missing
Screenshot 2020-06-05 at 16 44 09

SteRiccio:
Fixed #842


server/modules/category/categoryValidator.js
```diff
@@ -183,7 +183,11 @@ const validateItems = async (category, itemsByParentUuid) => {
     visitedUuids.add(itemUuid)
   }
 
-  return Validation.newInstance(R.isEmpty(itemsValidationsByUuid), itemsValidationsByUuid)
+  return R.isEmpty(itemsValidationsByUuid)
+    ? null
+    : Validation.newInstance(false, { ...itemsValidationsByUuid }, [
```
ramirobg94, line 188: why false?

SteRiccio: if there is at least one item validation object, then the items are invalid (only validation for invalid items is kept).
If the items are valid, we can return just null, otherwise a Validation object with valid set to false.

ramirobg94: thanks!

PR #1012, opened 2020-07-26
Bugfix: attribute not shown in entity table when added to another cycle

### LINKED ISSUE #945: FormDesigner: entity table doesn't add attributes in a different cycle
Steps:
Add entity table
Add some attributes
Add a new cycle
Edit the entity table by adding a new attribute only to cycle 1
Edit the newly added attribute by selecting cycle 2
Select cycle 2
Attribute is not visible under the parent entity table

SteRiccio:
Fixes #945


server/modules/nodeDef/manager/nodeDefManagerLayout.js
```diff
@@ -0,0 +1,145 @@
+import * as R from 'ramda'
+
```
minotogna, line 2: call this NodeDefLayoutManager

minotogna: can you add some comment to the methods below? It will make code and features more readable

SteRiccio: ok, sure

PR #1014, opened 2020-07-27
change fetches to use API

ramirobg94:
On this PR some fetches were replaced by the common API service.


webapp/components/survey/TaxonomyDetails/store/actions/useUpdate.js
```diff
@@ -15,17 +16,15 @@ export const useUpdate = ({ setState }) => {
...
     const action = async () => {
-      await axios.put(`/api/survey/${surveyId}/taxonomies/${Taxonomy.getUuid(taxonomy)}`, {
-        key,
-        value,
-      })
+      await API.updateTaxonomy({ surveyId, taxonomyUuid, data: { key, value } })
     }
     dispatch(SurveyActions.metaUpdated())
-    dispatch(debounceAction(action, Taxonomy.getUuid(taxonomy)))
+    dispatch(debounceAction(action, taxonomyUuid))
```
SteRiccio, line 28: should we use a more complex key when debounceAction is called, just to avoid collisions? something like `taxonomy_update_${taxonomyUuid}`?

ramirobg94: i dindt know, sure i will change :)

PR #1046, opened 2020-08-17
DataQuery - show aggregates (SUM)

### LINKED ISSUE #811: DataQuery: show aggregates SUM
Goal: show in table dimensions and measures aggregated using sum function.
When user select at least one dimension and one measure, data table shows by default measures aggregated using sum function

**Backend**
- [x] Api EndPoint: add a condition to invoke aggregate service function 

**FrontEnd DataQuery**
- [x] Invoke end point on selection update (use A.stringify to pass the query object to the server)
- [x] Result table must contain first the dimensions, then the measures.
- [x] Measure column headers should contain two rows (see for example code attributes in raw mode): first row the label of the measure, second row the capitalized aggregate function used (in this case SUM)

SteRiccio:
Resolves #811


server/modules/surveyRdb/repository/dataView/readAgg.js
```diff
@@ -12,8 +12,9 @@ import { Sort, Query } from '../../../../../common/model/query'
 const _getSelectFieldsMesaures = ({ survey, viewDataNodeDef, measures }) => {
   const fields = []
   const params = {}
-  Object.entries(measures).forEach(([nodeDefUuidMeasure, aggFunctions], i) => {
-    const paramName = `measure_field_${i}`
+  let index = 0
+  measures.forEach((aggFunctions, nodeDefUuidMeasure) => {
```
minotogna, line 16: the index is the 3rd argument of forEach. you don't need to declare it above

PR #1052, opened 2020-08-18
Bug/deselect category item

### LINKED ISSUE #1047: Dropdown: cannot deselect selected item
it can be useful for example in CodeProps / Parent code if the user selects a parent code node def by mistake and wants to clear the selected value.

- [ ] If user clear the value or current value doesn't match any valid item, selected item should be set to null

ramirobg94:
resolves #1047


webapp/components/form/Dropdown/store/actions/useBlurInputValue.js
```diff
@@ -0,0 +1,20 @@
+import { useCallback } from 'react'
+
+import * as A from '@core/arena'
+
+import { State } from '../state'
+import { getItemsDialog } from './getItemsDialog'
+
+export const useBlurInputValue = ({ onChange }) =>
```
SteRiccio, line 8: can this be just `useBlurInput` or `useOnBlurInput` and called from outside with `onBlurInput`?

minotogna (approved):

PR #1088, opened 2020-08-21
Attempt to solve issue with node def code items fetching

### LINKED ISSUE #1079: Error in console when adding rows in multiple entity with code attribute definitions
The error shown is: `Can't perform a React state update on an unmounted component.`

SteRiccio:
Fixes #1079

Main changes are in NodeDefCode.
A hook like useFetchItems or something similar could be introduced to move some of the code there.

### DIFF
--- a/webapp/components/survey/SurveyForm/nodeDefs/components/types/NodeDefCode/store/hooks/useItems.js
+++ b/webapp/components/survey/SurveyForm/nodeDefs/components/types/NodeDefCode/store/hooks/useItems.js
@@ -1,0 +1,34 @@
+import { useEffect, useRef, useState } from 'react'
+
+import * as API from '@webapp/service/api'
+
+export const useItems = ({ categoryUuid, categoryLevelIndex, draft, edit, parentCategoryItemUuid, surveyId }) => {
+  const [items, setItems] = useState([])
+  const cancelRequestRef = useRef(null)
+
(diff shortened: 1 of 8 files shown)


webapp/components/survey/SurveyForm/nodeDefs/components/types/NodeDefCode/store/hooks/useItems.js
```diff
@@ -0,0 +1,34 @@
...
+          : []
+      cancelRequestRef.current = cancel
+
+      fetchItems
+        .then(({ data: { items: itemsLoaded = {} } }) => {
+          setItems(itemsLoaded)
+        })
+        .catch(() => {
+          // canceled
+        })
```
ramirobg94, line 24: how about use async-await? also, the fetch on line 14 is async

ramirobg94 (changes requested):

ramirobg94 (approved):

webapp/components/survey/SurveyForm/nodeDefs/components/types/NodeDefCode/store/hooks/useItems.js
SteRiccio: we cannot "await" for the promise to resolve, otherwise we won't be able to cancel it when needed.

PR #1108, opened 2020-08-25
show real number of columns on designer tables

### LINKED ISSUE #1102: Designer with table: show true number of columns
In Designer with table there is count of columns as for a form. This number does not show number of columns in the table

image

It would be better to show true number of columns in this table.

ramirobg94:
Fixes #1102


ramirobg94:
This PR shouldnt be done"

PR #1131, opened 2020-08-27
E2etest/004 survey from edit cluster

ramirobg94:


test/e2e/utils/api/expect.js
```diff
@@ -1,10 +1,29 @@
 import { $, text as textTaiko } from 'taiko'
 
-export const expectExists = async ({ text = null, selector = null }) => {
-  if (!text && !selector) throw new Error('One between text and selector is required')
+export const getSelection = async ({ text = null, selector = null, itemSelector = null }) => {
```
minotogna, line 3: why selector and itemSelector? 
can you just pass the button as selector?
see where it's used, and use $ to be used as selector

test/e2e/utils/api/expect.js
```diff
@@ -1,10 +1,29 @@
...
 
   const exists = await element.exists()
   await expect(exists).toBeTruthy()
 }
+
+export const expectExistsExactlyNumberOfTimes = async ({
+  text = null,
+  selector = null,
+  itemSelector = null,
+  numberOfItems = 0,
```
minotogna, line 22: pass numberOfItems to expectExists. better not to have another method.
then default is 1. if you expect an item, at least it should be 1.
then update expectExists with await expect(items.length).toBe(numberOfItems)

minotogna (changes requested):

test/e2e/utils/api/expect.js
ramirobg94: removed

test/e2e/utils/api/expect.js
ramirobg94: changed

test/e2e/tests/004-surveyFormEditCluster.js
```diff
@@ -0,0 +1,81 @@
...
+  addItemToPage,
+  expectItemIsTheLastNodeDef,
+  expectItemsAreInOrderAsNodeDef,
+} from '../utils/ui'
+
+const expectSurveyFormLoaded = async () => expectExists({ selector: '.survey-form' })
+
+const expectHasOnlyRootEntity = async ({ rootEntityName }) => {
+  await expectExists({ selector: '.btn-node-def', numberOfItems: 1 })
+  await expectExists({ selector: `//button[text()='${rootEntityName}']`, numberOfItems: 1 })
```
minotogna, line 14: isn't numberOfItems by default 1? you can omit it here

test/e2e/tests/004-surveyFormEditCluster.js
```diff
@@ -0,0 +1,81 @@
...
+    await expectItemIsTheLastNodeDef({ item: child })
+  })
+
+  test('Cluster add children - verify order', async () => expectItemsAreInOrderAsNodeDef({ items: nodeDefItems }))
+
+  test('Cluster add children - verify number of children', async () =>
+    expectExists({
+      selector: '.survey-form__node-def-page-item',
+      numberOfItems: nodeDefItems.length,
+    }))
```
minotogna, line 80: does this only prove that there are `nodeDefItems.length` children? any practical way we can test the order as well?

test/e2e/utils/api/index.js
```diff
@@ -1,7 +1,7 @@
 export { closeBrowser, openBrowser } from './browser'
 export { button } from './button'
 export { click, clickParent } from './click'
-export { expectExists } from './expect'
+export { expectExists, getElement } from './expect'
```
minotogna, line 4: why is getElement exported from expect? 
add a new file element ?

test/e2e/utils/ui/surveyForm.js
```diff
@@ -0,0 +1,58 @@
+import { click, getElement, toRightOf, writeIntoTextBox } from '../api'
+import { waitForLoader } from './loader'
+
+const selectors = {
+  name: () => toRightOf('Name'),
+  label: () => toRightOf('Label'),
+  key: () => toRightOf('Key'),
+  multiple: () => toRightOf('Multiple'),
+}
```
minotogna, line 9: aren't these related to node def details? 
better to separate nodeDefDetails from surveyForm

minotogna (changes requested):

test/e2e/tests/004-surveyFormEditCluster.js
ramirobg94: I put one because exists is not the same as being an exact number of times for example  here root_entity is two times:

Captura de pantalla 2020-09-03 a las 11 21 01

If we like that something is on the screen it should exist and if we like to be sure that something is exactly one time on the screen we should check numberOfItems: 1.

What do you thing?

test/e2e/tests/004-surveyFormEditCluster.js
ramirobg94: I think that this comment is solved on line 74

test/e2e/utils/api/index.js
ramirobg94: agree 👍

test/e2e/utils/ui/surveyForm.js
ramirobg94: Perfect!

test/e2e/tests/004-surveyFormEditCluster.js
minotogna: I see now.
then I think we should have two methods.
one is expectToBeTruthy and another one expectToBe.
what do you think?

test/e2e/tests/004-surveyFormEditCluster.js
minotogna: sorry, missed that.

test/e2e/tests/004-surveyFormEditCluster.js
ramirobg94: How about
expectToBeTruthy -> is **expectExists**
expectToBe -> **expectToBe** ({ numberOfTimes })

ramirobg94: can I do the Refactor into the next PR 005 because I am going to have a lot of conflicts?

minotogna: sure

minotogna (approved):

PR #1135, opened 2020-08-28
NodeDefDetails: cancel edits when user presses browser back button

### LINKED ISSUE #1116: Error message after deleting a Code attribute
While adding a new CODE attribute, it is not clear to a user what to do after selecting a code list (=category list) item. There is X in the upper left corner, BUT IF user clicks browser's backwards arrow, Arena jumps into Form Designer View.
Then try to DELETE this attribute: there is this error message:
image

Selecting category requires a DONE button!

SteRiccio:
Fixes #1116

### DIFF
--- a/webapp/components/survey/NodeDefDetails/store/useNodeDefDetails.js
+++ b/webapp/components/survey/NodeDefDetails/store/useNodeDefDetails.js
@@ -1,4 +1,4 @@
-import { useEffect, useState } from 'react'
+import { useEffect, useRef, useState } from 'react'
 import { useHistory, useParams } from 'react-router'
 
 import { useIsDesignerNodeDefRoute, useOnUpdate } from '@webapp/components/hooks'
@@ -20,6 +20,7 @@ export const useNodeDefDetails = () => {
   const surveyCycleKey = useSurveyCycleKey()
 
   const [state, setState] = useState({})
+  const stateRef = useRef(state)
 
   const Actions = useActions({ setState })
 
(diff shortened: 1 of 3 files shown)


webapp/components/survey/NodeDefDetails/store/useNodeDefDetails.js
```diff
@@ -20,20 +20,33 @@ export const useNodeDefDetails = () => {
...
     // Editing a nodeDef
-    if (nodeDefUuid) {
+    if (nodeDefUuid && !State.getNodeDef(state)) {
       const nodeDefSurvey = Survey.getNodeDefByUuid(nodeDefUuid)(survey)
       const validation = Survey.getNodeDefValidation(nodeDefSurvey)(survey)
       setState(State.create({ nodeDef: nodeDefSurvey, validation }))
     }
+
+    return () => {
+      // get local state from stateRef to get it up to date in this callback
```
minotogna, line 38: what does that mean?

webapp/components/survey/NodeDefDetails/store/useNodeDefDetails.js
```diff
@@ -20,20 +20,33 @@ export const useNodeDefDetails = () => {
...
+      if (State.isDirty(stateRef.current)) {
+        Actions.cancelEdits({ state: stateRef.current, showConfirm: false, goBack: false })
+      }
+    }
   }, [])
 
+  useEffect(() => {
+    // update state ref on local state update
+    stateRef.current = state
+  }, [state])
```
minotogna, line 48: why keeping a ref to the local state and updating it every time the state changes?

minotogna (changes requested):

webapp/components/survey/NodeDefDetails/store/useNodeDefDetails.js
SteRiccio: the local state that you get from the "state" variable is no up to date in this callback.
The only way I've found to get an updated state there was to put it in a "ref".

webapp/components/survey/NodeDefDetails/store/useNodeDefDetails.js
SteRiccio: see comment to line 36

webapp/components/survey/NodeDefDetails/store/useNodeDefDetails.js
minotogna: of course. have you tried to pass the state to the hook input param?

SteRiccio: ...the cancelEdits function should be called only on unmount, not every time the state changes...

minotogna: Do you necessarily need to pass the state to that action ?
Keeping a ref that is just a reference to the state looks a bit dirty to me

PR #1137, opened 2020-08-28
Fixed min/max count errors not shown in code attributes

### LINKED ISSUE #1134: Multiple code attribute: Minimum and Maximum required number, and Checkbox list
In the following example I have 2 multiple code attributes on the right side of UI, using the same categorical list:
image

in both cases there is property Mininum set to 0, and Maximum items to 2. This validation rule is not taken into account yet as an Error.

In addition, neither of input objects look good. Please add Checkbox list into the system, as follows:
image

SteRiccio:
Fixes #1134 
The change that fixes the issue is only in the file webapp/components/survey/SurveyForm/nodeDefs/components/useValidation.js line 31.
The other changes have been done only to avoid ESLint errors.

### DIFF
--- a/core/record/_recordValidator/countValidator.js
+++ b/core/record/_recordValidator/countValidator.js
@@ -10,29 +10,37 @@ import * as NumberUtils from '@core/numberUtils'
 import * as Record from '../record'
 import * as Node from '../node'
 
+const _createValidationResult = (nodeDefUuid, isMinCountValidation, minCount, maxCount) =>
+  Validation.newInstance(false, {}, [
+    {
+      key: isMinCountValidation
+        ? Validation.messageKeys.record.nodesMinCountNotReached
+        : Validation.messageKeys.record.nodesMaxCountExceeded,
+      params: {
+        nodeDefUuid,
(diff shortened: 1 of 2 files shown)


core/record/_recordValidator/countValidator.js
```diff
@@ -47,63 +55,50 @@ const _getNodePointers = (survey, record, nodes) => {
     if (NodeDef.isEntity(nodeDef) && !Node.isDeleted(node)) {
       // Add children node pointers
       const childDefs = Survey.getNodeDefChildren(nodeDef)(survey)
-
-      for (const childDef of childDefs) {
-        nodePointers.push({
+      nodePointers.push(
+        ...childDefs.map((childDef) => ({
```
ramirobg94, line 59: is it always an array?

ramirobg94 (approved):

core/record/_recordValidator/countValidator.js
SteRiccio: childDefs is returned by Survey.getNodeDefChildren, and it's an array

PR #1138, opened 2020-08-31
Show confirm dialog when user presses back button and node def is being edited

SteRiccio:


webapp/components/survey/NodeDefDetails/store/actions/useCancelEdits.js
```diff
@@ -7,30 +7,44 @@ import { NodeDefsActions } from '@webapp/store/survey'
...
-export const useCancelEdits = () => {
+export const useCancelEdits = ({ setState }) => {
   const history = useHistory()
   const dispatch = useDispatch()
 
   const cancelEdits = ({ state }) => async () => {
     const nodeDef = State.getNodeDef(state)
     const nodeDefOriginal = State.getNodeDefOriginal(state)
 
+    await setState(State.assocNodeDef(nodeDefOriginal))
```
ramirobg94, line 18: do we like to restore to the original node def? or just remain the node dirty?

webapp/components/survey/NodeDefDetails/store/actions/useCancelEdits.js
```diff
@@ -7,30 +7,44 @@ import { NodeDefsActions } from '@webapp/store/survey'
...
-          key: 'common.cancelConfirm',
-          onOk: cancelEdits({ state }),
-        })
-      )
-    } else {
-      await cancelEdits({ state })()
-    }
-  }, [])
+  return useCallback(
+    ({ state }) =>
```
ramirobg94, line 26: do we need to add async?

SteRiccio: Uhm, yes, so it's clear it returns a Promise

webapp/components/survey/NodeDefDetails/store/actions/useCancelEdits.js
SteRiccio: this is necessary to "disable" the interception of back button press when user presses the Cancel button and history.back() is called

core/i18n/resources/en.js
```diff
@@ -57,6 +57,7 @@ Do you want to cancel them?`,
     label: 'Label',
     label_plural: 'Labels',
     language: 'Language',
+    leavePageConfirmMessage: 'Data will be lost if you leave the page, are you sure?',
```
minotogna, line 60: better to say:
`There are unsaved changes in the form.
By confirming all changes will be lost.
Do you want to proceed?`

webapp/components/hooks/useInterceptBack.js
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * @param {!object} params - The parameters.
+ * @param {!Function} [params.onBack] - Callback to be called when back button is pressed. It should return true when history.back() is called by the callback itself, false othwerwise.
+ * @param {boolean} [params.active=false] - True if onBack callback should be called on browser back button press and confirm message on page refresh should be shown.
+ * @returns {undefined}
+ */
+export const useInterceptBack = (params) => {
+  const { active = false, onBack } = params
+
+  const onBackButtonEvent = async (e) => {
```
minotogna, line 15: call e event

webapp/components/hooks/useInterceptBack.js
```diff
@@ -0,0 +1,39 @@
...
+    e.preventDefault()
+
```
minotogna, line 17: why using window.history instead of react-router history ?

webapp/components/hooks/usePageUnloadConfirm.js
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * Intercepts the page unload event and shows a confirm message, if active.
+ *
+ * @param {object} [params={}] - The parameters.
+ * @param {boolean} [params.active=false] - True if a message on page unload should be shown, false otherwise.
+ * @param {string} [params.confirmMessageKey=common.leavePageConfirmMessage] - Key of the leave page confirm message.
+ * @returns {undefined}
+ */
+export const usePageUnloadConfirm = (params = {}) => {
```
minotogna, line 13: call this useOnPageUnload

webapp/components/hooks/useInterceptBack.js
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * Intercepts the browser back button press and calls the specified callback.
+ * The call to the callback can be can be activated or deactivated using the "active" parameter.
+ *
+ * @param {!object} params - The parameters.
+ * @param {!Function} [params.onBack] - Callback to be called when back button is pressed. It should return true when history.back() is called by the callback itself, false othwerwise.
+ * @param {boolean} [params.active=false] - True if onBack callback should be called on browser back button press and confirm message on page refresh should be shown.
+ * @returns {undefined}
+ */
+export const useInterceptBack = (params) => {
```
minotogna, line 12: call this useOnBrowserBack

webapp/components/survey/NodeDefDetails/store/actions/useCancelEdits.js
```diff
@@ -7,30 +7,44 @@ import { NodeDefsActions } from '@webapp/store/survey'
...
+              },
+              onCancel: () => {
+                resolve(false)
+              },
+            })
+          )
+        } else {
+          cancelEdits({ state })()
+            .then(() => resolve(true))
+            .catch((e) => reject(e))
```
minotogna, line 45: not possible to use async/await ?

webapp/components/survey/NodeDefDetails/store/actions/useCancelEdits.js
```diff
@@ -7,30 +7,44 @@ import { NodeDefsActions } from '@webapp/store/survey'
...
-          onOk: cancelEdits({ state }),
-        })
-      )
-    } else {
-      await cancelEdits({ state })()
-    }
-  }, [])
+  return useCallback(
+    async ({ state }) =>
+      new Promise((resolve, reject) => {
```
minotogna, line 27: is there a need to return new Promise here?

minotogna (changes requested):
good idea these 2 hooks

webapp/components/hooks/useInterceptBack.js
```diff
@@ -0,0 +1,39 @@
...
+      if (!wentBack) {
+        window.history.pushState(null, null, window.location.pathname)
+      }
+    } else {
+      window.history.back()
+    }
+  }
+
+  useEffect(() => {
+    // Push a "null" state to detect back button press
```
ramirobg94, line 29: Take a look of this link, the note is interesting:
https://developer.mozilla.org/en-US/docs/Web/API/History_API/Working_with_the_History_API

https://developer.mozilla.org/en-US/docs/Web/API/History/pushState

core/i18n/resources/en.js
SteRiccio: ok

webapp/components/hooks/useInterceptBack.js
SteRiccio: ok

webapp/components/hooks/useInterceptBack.js
SteRiccio: how?

webapp/components/hooks/usePageUnloadConfirm.js
SteRiccio: ok

webapp/components/survey/NodeDefDetails/store/actions/useCancelEdits.js
SteRiccio: since the Promise can resolve inside the onOk or onCancel callbacks passed to showDialogConfirm, the only way to resolve it there was by using its methods `then` and `catch`.

webapp/components/survey/NodeDefDetails/store/actions/useCancelEdits.js
SteRiccio: see comment to line 43

webapp/components/hooks/useInterceptBack.js
SteRiccio: ok

webapp/components/survey/NodeDefDetails/store/actions/useCancelEdits.js
minotogna: can't you add a try/catch instead ?

SteRiccio: I've tried... but there is a problem: `Promise executor functions should not be async`
So, in the end this solution keeps the code a bit more readable.

webapp/components/hooks/useInterceptBack.js
minotogna: useHistory ?

SteRiccio: useHistory... useHistoryListen... you can only listen to changes to the location. The best thing would be to avoid changes in the location, so you don't unmount/mount components.

minotogna: I meant to use history api https://reactrouter.com/web/api/history

SteRiccio: ok

minotogna (approved):

PR #1139, opened 2020-09-04
Webpack: Add hot reloading

sorja:
Adds hot reloading for frontend, caching for server (faster reloads/builds : 6s => 2s)
Note: Also adds morgan for server side request logging (on dev, local)


webpack.config.babel.js
```diff
@@ -67,7 +72,7 @@ const plugins = [
     filename: fontCssFileName,
   }),
   new CleanUpStatsPlugin(),
-]
+].filter(Boolean)
```
SteRiccio, line 75: it's not very clear... can you push the development plugins in a "if" block instead?

server/system/appCluster.js
```diff
@@ -31,6 +33,10 @@ export const run = async () => {
 
   const app = express()
 
+  if (ProcessUtils.ENV.nodeEnv === 'development') {
```
SteRiccio, line 36: you can use ProcessUtils.isEnvDevelopment instead

SteRiccio (changes requested):

webpack.config.babel.js
```diff
@@ -93,17 +101,50 @@ const webPackConfig = {
...
+            loader: 'babel-loader',
+            options: {
+              plugins: isDevelopment ? [require.resolve('react-refresh/babel')] : [],
+            },
+          },
+        ],
       },
       {
-        test: /\.(sa|sc|c)ss$/,
+        test: /\.(sass|scss|css)$/,
```
minotogna, line 147: was this change necessary ?

SteRiccio: I don't know if it's necessary or not but to me it looks more clear in this way

ramirobg94: For me too 👍

ramirobg94 (approved):

webpack.config.babel.js
```diff
@@ -93,17 +101,50 @@ const webPackConfig = {
...
     publicPath: '/',
   },
+  devServer: {
+    hot: true,
+    index: '',
+    proxy: [
+      {
+        // Proxy all server-served routes:
+        context: ['/img', '/api', '/auth', '/socket.io', 'sockjs-node'],
+        target: 'http://localhost:9090',
```
ramirobg94, line 111: should this be moved to some var?

webpack.config.babel.js
sorja: For grepping

webpack.config.babel.js
sorja: the target?

sorja: I think it's clear like this - it is not used twice

minotogna (approved):

minotogna:
can we merge this branch?

SteRiccio (approved):

PR #1152, opened 2020-09-11
expressions: Added ln, log10, exp (**)

### LINKED ISSUE #1119: Expression editor: Add Exponentiation operator (**)
We need exponential operator in UI, and preferably Math.log() function that is equivalent to ln(x) in mathematics. These both two are really commonly used in forest inventory applications.
Also Math.log10 is useful!
We also need a function to check whether object length is 0 (zero). This was easy in Xpath.

sorja:
Resolves #1119 

Note: Does not add length0 check


minotogna (changes requested):
don't we handle this with pow already ?

sorja:
> 
> 
> don't we handle this with pow already ?

Yes, we indeed have pow. But I added also exp (**) as an exercise how things work

minotogna (approved):

PR #1153, opened 2020-09-11
cancel request on unmount

### LINKED ISSUE #1008: ActivilyLog unmount warnings
`Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.
    in ActivityLog (created by Dashboard)
    in div (created by Dashboard)
    in SurveyDefsLoader (created by Dashboard)
    in Dashboard (created by Context.Consumer)
    in Route (created by ModuleSwitch)
`

ramirobg94:
fixes #1008


minotogna (changes requested):
a cancelable axios was already implemented. 
please check with @SteRiccio to use only one way.

webapp/service/api/cancelableRequest.js
```diff
@@ -0,0 +1,13 @@
+import axios from 'axios'
+
+export const cancelableRequest = ({ method = 'get', url, config = {} }) => {
```
SteRiccio, line 3: I wouldn't pass the entire config object but only the "params"

SteRiccio (approved):

webapp/service/api/cancelableRequest.js
minotogna: params is only used by get, data is used by post.
if you want to pass only params, then we need a canelableGetRequest that calls the generic cancelableRequest

minotogna (changes requested):

webapp/service/api/cancelableRequest.js
SteRiccio: can make the interface homogeneous among get and post, pass only params and take care of how the config object is created only inside the cancelableRequest function? it would simplify a bit the usage of it.

SteRiccio: to me an homogeneous interface looks nicer and if for some reason you have to switch from a get to a post it would be easier.
Passing the entire "config" object makes it dependent on the implementation (axios) and it's not true that developers are more used in using different ways to send a get or a post; consider for example the JQuery shorthands $.get and $.post, they both take a "data" object as parameter and they are just shorthands for $.ajax({url, type, data, ...})

minotogna (approved):

PR #1162, opened 2020-09-17
Fixed update node request sent to server when default value is applied

### LINKED ISSUE #1161: Meta attribute defaultValue not set to true when default value is applied
The meta attribute defaultValue is set to false while it should be set to true when the default value is applied.
This is due to a node update request sent to the server when the value of a readOnly attribute is updated.

SteRiccio:
Fixes #1161


webapp/components/form/Input/Input.js
```diff
@@ -32,7 +34,8 @@ export const Input = React.forwardRef((props, ref) => {
 
   const handleValueChange = (newValue) => {
```
SteRiccio, line 35: this function is called by the NumberFormat input field even when the value is not changed by the user, but it changes in the passed props.
To avoid a useless (and dangerous) call to onChange, we can check that the formatted value (newValue) is different from the value coming from the props.

sorja (approved):

core/stringUtils.js
```diff
@@ -29,6 +29,15 @@ export const isNotBlank = R.pipe(isBlank, R.not)
...
 export const removeNewLines = R.when(isString, R.pipe(R.split(/\r\n|\r|\n/g), R.join(' ')))
+
+/**
+ * Returns an empty string if the passed value is null or it's an invalid number,
+ * otherwise returns the specified value converted to string.
+ *
+ * @param {!any} value
+ * @returns {string} Empty string if the specified value is null or an invalid number or the specified value.
+ */
+export const nullToEmpty = (value) => (value === null || Number.isNaN(value) ? '' : String(value))
```
minotogna, line 43: isn't this related to numbers ?
to me it looks like a NumberUtils.toString

minotogna (changes requested):

minotogna (approved):

PR #1223, opened 2020-11-04
Fixed validation of expressions using attributes inside single entities

### LINKED ISSUE #702: Expression editor: Allow usage of single entities attributes children of nodeDef context hierarchy
I imported a survey "1- BIOPHYSICAL SURVEY - CIRCULAR PLOT WITH MULTIPLE ENTITIES"
and in Collect Import Report there is line as follows:
Path: Cluster > Plots > Canopy Closure
Applicable if: plot_land_use/land_use='1'

In Expression Editor, there is no Land Use shown.

SteRiccio:
Fixes #702

### DIFF
--- a/core/survey/_surveyValidator/nodeDefExpressionsValidator.js
+++ b/core/survey/_surveyValidator/nodeDefExpressionsValidator.js
@@ -9,7 +9,7 @@ import * as NodeDefValidations from '@core/survey/nodeDefValidations'
 import * as NodeDefExpression from '@core/survey/nodeDefExpression'
 import * as Expression from '@core/expressionParser/expression'
 import * as ObjectUtils from '@core/objectUtils'
-
+import Queue from '@core/queue'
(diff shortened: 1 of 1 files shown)


core/survey/_surveyValidator/nodeDefExpressionsValidator.js
```diff
@@ -116,7 +134,7 @@ const _validateNodeDefExpr = async (survey, nodeDef, dependencyType, expr) => {
 
 const _validateExpressionProp = (survey, nodeDef, dependencyType) => async (propName, item) => {
```
ramirobg94, line 135: is this async needed?

SteRiccio: yes because it call _validateNodeDefExpr that is async!

ramirobg94 (approved):

PR #1234, opened 2020-11-10
Bugfix/collect import specify field

### LINKED ISSUE #1192: Collect import hides attribute (Specify for multientry code in a table)
Collect import does not show SPECIFY text attributes in a table, if the relevancy is linked to multi-entry code attribute

this is case in Collect survey:
image

and this error when trying to publish in Arena:
image

and this is view in Arena editor:
image

SteRiccio:
Fixes #1192

### DIFF
--- a/core/expressionParser/expression.js
+++ b/core/expressionParser/expression.js
@@ -53,7 +53,7 @@ const isType = (type) => R.propEq('type', type)
 // Return true if the nodeDef can be used in expressions and false otherwise
 export const isValidExpressionType = (nodeDef) =>
   !NodeDef.isEntity(nodeDef) &&
-  !NodeDef.isMultiple(nodeDef) &&
+  //!NodeDef.isMultiple(nodeDef) &&
   !NodeDef.isCoordinate(nodeDef) &&
   !NodeDef.isFile(nodeDef)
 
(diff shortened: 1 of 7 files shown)


core/expressionParser/expression.js
```diff
@@ -53,7 +53,7 @@ const isType = (type) => R.propEq('type', type)
 // Return true if the nodeDef can be used in expressions and false otherwise
 export const isValidExpressionType = (nodeDef) =>
   !NodeDef.isEntity(nodeDef) &&
-  !NodeDef.isMultiple(nodeDef) &&
+  //!NodeDef.isMultiple(nodeDef) &&
```
sorja, line 56: Should this commented line be removed ?

core/expressionParser/helpers/evaluator.js
```diff
@@ -19,6 +19,7 @@ const stdlib = {
   // arity 1+ (arity 0 allowed by JS)
   min: [Math.min, 1, -1],
   max: [Math.max, 1, -1],
+  includes: [(items, value) => items && items instanceof Array && items.includes(String(value)), 2, 2],
```
sorja, line 22: should use `Array.isArray(items)` replacing ` items && items instanceof Array`

minotogna (approved):

core/expressionParser/expression.js
SteRiccio: yes sure!

core/expressionParser/helpers/evaluator.js
SteRiccio: yes, thanks, you are right

minotogna (approved):

PR #1239, opened 2020-11-11
Paginator: replace current entry in history stack when changing page

### LINKED ISSUE #1059: Taxonomy editor: DONE button scrolls list backwards
When in a multipage taxonomic list user goes to page 2 or forwards, DONE button scrolls back the previous page and does not return the view of all taxonomic lists as it should do.
It is working wrong now.

SteRiccio:
Fixes #1059

### DIFF
--- a/webapp/components/Table/Header/Header.js
+++ b/webapp/components/Table/Header/Header.js
@@ -21,7 +21,7 @@ const Header = (props) => {
           offset={offset}
           limit={limit}
           count={count}
-          setOffset={(offsetUpdated) => history.push(getLink(offsetUpdated))}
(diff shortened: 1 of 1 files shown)


webapp/components/Table/Header/Header.js
```diff
@@ -21,7 +21,7 @@ const Header = (props) => {
           offset={offset}
           limit={limit}
           count={count}
-          setOffset={(offsetUpdated) => history.push(getLink(offsetUpdated))}
+          setOffset={(offsetUpdated) => history.replace(getLink(offsetUpdated))}
```
minotogna, line 24: what if you are in survey list and go back?

minotogna (approved):

PR #1251, opened 2020-11-13
Capitalize survey name on header

ramirobg94:
Small fix


webapp/views/App/Header/Header.scss
```diff
@@ -49,6 +49,7 @@
 
     &__name {
       font-size: 0.6rem;
+      text-transform: capitalize;
```
minotogna, line 52: Do not capitalize the survey name .

minotogna (changes requested):

PR #1261, opened 2020-11-15
1248-dropdown-list-overflow-bottom-of-page

### LINKED ISSUE #1248: Dropdown list extends past bottom of the page
When a dropdown list is at the very end of the page, it will go through the bottom of the page, see this example
image

image

ramirobg94:
fixes #1248


SteRiccio (approved):

webapp/components/form/AutocompleteDialog/store/state/state.js
```diff
@@ -24,12 +26,19 @@ export const getList = A.prop(keys.list)
 
 export const getItemsSize = (state) => getItems(state).length
 
+const _calculateVerticalPosition = ({ top: elementTop, height: elementHeight }) => {
+  const top = elementTop + elementHeight
+  const windowScreenY = window.innerHeight
+
+  return top + MAX_HEIGHT_AUTOCOMPLETE_DIAOLOG > windowScreenY ? { bottom: windowScreenY - elementTop } : { top }
+}
```
minotogna, line 34: does this open the dropdown on top?
if yes, then it shouldn't. update body height

minotogna (changes requested):

webapp/components/form/AutocompleteDialog/store/state/state.js
minotogna: because users aren't used to that

minotogna (approved):

PR #1262, opened 2020-11-15
1237-remove-subpage-breaks-editor

### LINKED ISSUE #1237: Remove a Subpage in Form editor leaves the app inconsistent
ezgif com-gif-maker (1)

ramirobg94:
Fixes #1237

### DIFF
--- a/webapp/components/survey/SurveyForm/components/FormPageNavigation/store/state/state.js
+++ b/webapp/components/survey/SurveyForm/components/FormPageNavigation/store/state/state.js
@@ -17,6 +17,7 @@ const keys = {
   expandedFormPageNavigation: 'expandedFormPageNavigation',
   canEditDef: 'canEditDef',
   surveyCycleKey: 'surveyCycleKey',
+  outerPageChildDefs: 'outerPageChildDefs',
 }
 
 // ===== CREATE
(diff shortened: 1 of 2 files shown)


webapp/components/survey/SurveyForm/components/FormPageNavigation/store/state/state.js
```diff
@@ -17,6 +17,7 @@ const keys = {
   expandedFormPageNavigation: 'expandedFormPageNavigation',
   canEditDef: 'canEditDef',
   surveyCycleKey: 'surveyCycleKey',
+  outerPageChildDefs: 'outerPageChildDefs',
```
minotogna, line 20: where do you read those? from what I see, the formPageNavigavtion always filters the nodeDef children.
isn't the problem somewhere when removing the nodeDef, i.e. the nodeDef is not removed from the store?

ramirobg94: The getter was declared on line 61 in this file, the key was missing and I added. 
Also, this getter is used in `arena/webapp/components/survey/SurveyForm/components/FormPageNavigation/FormPageNavigation.js`
line 18

The noDef was removed from the store but the problem was that we are not updating the local state, this was handled in the other file.

minotogna (approved):

PR #1285, opened 2020-11-19
Feature/form header path

### LINKED ISSUE #1133: Data-Records: show key attributes' values of parents
While editing attribute of child entity (as plot under a cluster), there should be visible the key attribute values for parents, in this case cluster's key attributes. Otherwise I do not know (after a coffee break ;) ) that which cluster data I am editing or adding in the system.

An example: while entering tree table data for a plot section, I should see above that this tree belongs to:
cluster_id: 24A
plot_no: 3
plot_section_id: A

minotogna:
Resolves #1133


webapp/components/survey/NodeDefLabelSwitch/NodeDefLabelSwitch.js
```diff
@@ -5,30 +5,27 @@ import * as NodeDef from '@core/survey/nodeDef'
 
 import { useI18n } from '@webapp/store/system'
 
-const NodeDefLabelSwitch = ({ onChange, labelType }) => {
+const NodeDefLabelSwitch = (props) => {
+  const { className, onChange, labelType } = props
   const i18n = useI18n()
 
+  const label = labelType === NodeDef.NodeDefLabelTypes.label ? 'common.showNames' : 'common.showLabels'
```
SteRiccio, line 12: this one should be called labelKey

SteRiccio (dismissed):

webapp/components/survey/NodeDefLabelSwitch/NodeDefLabelSwitch.js
minotogna: yeah it will... next PR

SteRiccio (approved):

SteRiccio (approved):

PR #1289, opened 2020-11-20
Fixed error importing Collect survey categories (do not camelize nested props)

### LINKED ISSUE #1266: Collect import crashes: Two files for data import
Importing this survey causes a crash:
https://drive.google.com/file/d/1jZ7j6TiiavZjBVIA8PdPbS42p77NoGHF/view?usp=sharing

image

In addition. This survey crashes at Taxonomic data import stage:
https://drive.google.com/file/d/1jZ7j6TiiavZjBVIA8PdPbS42p77NoGHF/view?usp=sharing

image

SteRiccio:
Fixes #1266
The fix is small but can have side effects on the whole application.
As far as I have seen, everything seems to work fine.
Since the amount code of camelize is very small, should we create our own A.camelize function that accepts a list of properties to skip? Or should we suggest this change to the developer and open a pull request?! I vote for the 2nd...

### DIFF
--- a/core/arena/camelize.js
+++ b/core/arena/camelize.js
@@ -1,0 +1,31 @@
+const _camelCase = (str) => str.replace(/[_.-](\w|$)/g, (_, x) => x.toUpperCase())
+
+const _walk = ({ object, skip = [] }) => {
+  if (!object || !(object instanceof Object) || object instanceof Date || object instanceof RegExp) {
+    return object
+  }
+  if (Array.isArray(object)) {
+    return object.map(_walk)
+  }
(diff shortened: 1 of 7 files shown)


minotogna:
> Fixes #1266
> The fix is small but can have side effects on the whole application.
> As far as I have seen, everything seems to work fine.
> Since the amount code of camelize is very small, should we create our own A.camelize function that accepts a list of properties to skip? Or should we suggest this change to the developer and open a pull request?! I vote for the 2nd...

I doubt camelize must change.
Better to implement `A.camelize = ({skip:[]})`

minotogna (approved):

ramirobg94 (approved):

PR #1291, opened 2020-11-22
Feature/expression editor group attributes by parent

### LINKED ISSUE #1129: ExpressionEditor: group attributes by parent entity
in Editor attributes are seen as a long list. I would be easier to use them if they are grouped by parent entity, and they become to the list in the same order as they they are in the entry form or in table. Now items are in alphabetic order (by parent entities). 
This is a very good example what we need:
see "Grouping with Category" at https://www.syncfusion.com/javascript-ui-controls/js-dropdown-list

ramirobg94:
resolves #1129 

The PR looks long, but:

- there are line 160 lines on the yarn lock
- There are too many lines changed by eslint and prettier

The changes are:

`webapp/components/expression/expressionVariables.js`
getVariablesGroupedByParentUuid -> function to group by parentUuid
getVariables -> some changes to avoid repetition and populate values with parent uuid

`webapp/components/expression/nodes/identifier.js`
replace Dropdown by Select

`webapp/components/form/Select/*`
Select with react-select and config files


webapp/components/form/Select/Select.js
```diff
@@ -0,0 +1,55 @@
...
+  customStyle: PropTypes.object,
+  className: PropTypes.string,
+  components: PropTypes.object,
+  options: PropTypes.array,
+  value: PropTypes.object,
+  onChange: PropTypes.func,
+}
+
+Select.defaultProps = {
+  customStyle: {},
```
minotogna, line 47: call this style

webapp/style/_vars.scss
```diff
@@ -23,6 +23,28 @@ $black: #2f3138;
...
+  --blueLight: #c5dbf0;
+  --blueLightFocus: #edf4fa;
+  --blueLightActive: #9dc3e5;
+  --blueDark: #265e8f;
+  --orange: #eba444;
+  --orangeLight: #f9ccb2;
+  --black: #2f3138;
+  --yellowHighlight: #f8f6cc;
+  --white: #ffffff;
+}
```
minotogna, line 46: this doesn't go here. to main style, also use the vars instead of explicit colors

minotogna (changes requested):
why passing another object called variablesGroupedByParentUuid and variables?
can't you just pass a boolean to the ExpressionEditor and then one set of variables to inner components?

webapp/components/expression/expressionVariables.js
```diff
@@ -69,24 +71,55 @@ const getChildDefVariables = ({ survey, nodeDefContext, nodeDefCurrent, mode, la
...
+    }),
+    {}
+  )
+
+  const variablesGrouped = Object.keys(variablesGroupedByParentUuid)
+    .map((parentUuid) => {
+      const parentNodeDef = Survey.getNodeDefByUuid(parentUuid)(survey)
+      return { parentUuid, label: NodeDef.getName(parentNodeDef), id: Number(NodeDef.getId(parentNodeDef)) }
+    })
+    .sort((groupA, groupB) => (groupA.id > groupB.id ? 1 : -1))
```
SteRiccio, line 94: why are they sorted by id? should they be sorted by label or name or by hierarchy level instead? I prefer the hierarchy level

webapp/components/expression/expressionVariables.js
```diff
@@ -69,24 +71,55 @@ const getChildDefVariables = ({ survey, nodeDefContext, nodeDefCurrent, mode, la
...
+  variables = variables.filter((variable) => {
+    if (variablesUuids.includes(variable.uuid)) return false
+    variablesUuids.push(variable.uuid)
+    return true
+  })
+
   // Show current node def variable in the first position
   const nodeDefCurrentUuid = NodeDef.getUuid(nodeDefCurrent)
-  variables.sort((varA, varB) => {
+  variables = variables.sort((varA, varB) => {
```
SteRiccio, line 122: sort should do side effect...

webapp/components/form/Select/Select.js
```diff
@@ -0,0 +1,55 @@
...
+  customStyle = {},
+  className: customClassName,
+  components: overrideComponents,
+  options,
+  value,
+  onChange,
+  ...props
+}) => {
+  return (
+    <>
```
SteRiccio, line 20: the fragment is not needed

webapp/style/_vars.scss
ramirobg94: Using the Sass var the value appears null 
Captura de pantalla 2020-11-23 a las 10 29 13
Captura de pantalla 2020-11-23 a las 10 29 09

webapp/components/form/Select/Select.js
ramirobg94: fixed

webapp/components/form/Select/Select.js
ramirobg94: Thanks at the beginning I had the Dropdown to compare

webapp/components/expression/expressionVariables.js
ramirobg94: fixed

webapp/style/_vars.scss
minotogna: fix it then.
you could eventually add these to a new sass file (cssVars)

ramirobg94: Solved!

webapp/components/expression/expressionVariables.js
ramirobg94: The same order as the Form is required, if we use hierarchy this level is not going to be the same, I am not sure if it works in the future with the ID, do we have any way to get the order?

minotogna: the order in the form is given by layout prop, not IDs.

webapp/components/expression/expressionEditorPopupState.js
```diff
@@ -106,6 +116,7 @@ export const useExpressionEditorPopupState = (props) => {
   const nodeDefCurrent = nodeDefUuidCurrent ? Survey.getNodeDefByUuid(nodeDefUuidCurrent)(survey) : null
 
   const variables = ExpressionVariables.getVariables({ survey, nodeDefContext, nodeDefCurrent, mode, lang })
+  const variablesGroupedByParentUuid = ExpressionVariables.getVariablesGroupedByParentUuid({ variables, survey })
```
minotogna, line 119: why two set of variables? wouldnt passing a boolean prop to indicate whether the variables should appear grouped by parent entity be enough?

webapp/components/form/Select/Select.js
```diff
@@ -0,0 +1,53 @@
...
+import { adaptSelection } from './utils'
+
+export const Select = ({
+  style = {},
+  className: customClassName,
+  components: overrideComponents,
+  options,
+  value,
+  onChange,
+  ...props
```
minotogna, line 17: do not spread props

webapp/components/form/Select/Select.js
```diff
@@ -0,0 +1,53 @@
...
+      className={customClassName}
+      classNamePrefix="select"
+      options={options}
+      onChange={onChange}
+      value={adaptSelection(value)}
+      isSearchable
+      theme={baseTheme}
+      components={{ ...defaultComponents, ...overrideComponents }}
+      styles={{ ...baseStyles, ...style }}
+      {...props}
```
minotogna, line 30: do not spread props

webapp/components/expression/expressionEditorPopupState.js
```diff
@@ -18,7 +18,17 @@ const initialState = {
...
+  const {
+    canBeConstant,
+    expr,
+    mode,
+    type,
+    nodeDefUuidContext,
+    nodeDefUuidCurrent,
+    onChange,
+    query,
+    isGroupedByParentEntity = true,
```
minotogna, line 30: can we shorten this name (and everywhere it's called this way)? groupByParent ?

webapp/components/expression/nodes/identifier.js
```diff
@@ -1,16 +1,41 @@
...
-    onChange={item => onChange(R.assoc('name', R.propOr('', 'value', item), node))}
-  />
-)
+import Select from '@webapp/components/form/Select'
+
+const findValue = ({ variables, node }) => {
+  let value = null
+  value = R.find(R.propEq('value', node.name), variables)
+  if (!A.isEmpty(value)) return value
+  const _flatVariables = variables.reduce((flat, group) => [...(flat || []), ...(group.options || [])], [])
```
minotogna, line 13: can flat be null? ...(flat || [])  probably not.
also, give meaningful names. what is flat? variablesAggregator?

webapp/components/form/Select/Select.js
```diff
@@ -0,0 +1,51 @@
...
+import './Select.scss'
+import React from 'react'
+import PropTypes from 'prop-types'
+import ReactSelect from 'react-select'
+
+import { defaultComponents, baseStyles, baseTheme } from './config'
+import { adaptSelection } from './utils'
+
+export const Select = ({
+  style = {},
```
minotogna, line 11: you already assing {} in defaultProps

webapp/components/form/Select/Select.js
```diff
@@ -0,0 +1,51 @@
...
+  className: customClassName,
```
minotogna, line 12: any reason to rename this to customClassName ?

webapp/components/form/Select/Select.js
```diff
@@ -0,0 +1,51 @@
...
+  components: overrideComponents,
```
minotogna, line 13: any reason to rename this to overrideComponents ?

minotogna (changes requested):

webapp/components/form/Select/Select.js
ramirobg94: not really

webapp/components/form/Select/Select.js
ramirobg94: not just I used to do that when I use styled-components

webapp/components/form/Select/Select.js
ramirobg94: Thanks!

webapp/components/expression/nodes/identifier.js
ramirobg94: Changed!

webapp/components/expression/expressionEditorPopupState.js
ramirobg94: Done

webapp/components/expression/expressionVariables.js
ramirobg94: Not sure about how to sort elements using these layout props, do we have any method to have the position?

webapp/components/form/Select/Select.js
```diff
@@ -0,0 +1,42 @@
+/* eslint-disable react/jsx-props-no-spreading */
```
minotogna, line 1: I'm sure you can get rid of that eslint-disable now

webapp/components/expression/expressionVariables.js
minotogna: we dont have a method now. you could eventually implement it now, or we'll handle the sorting separately later on.
up to you, but let us know so that we can create an issue in case we'll do it later

webapp/components/form/Select/Select.js
ramirobg94: Thanks

webapp/components/expression/expressionVariables.js
ramirobg94: I added a method,
not sure if the name is good enough.
I will leave a comment there

webapp/components/expression/expressionVariables.js
```diff
@@ -75,6 +77,20 @@ const getChildDefVariables = ({ survey, nodeDefContext, nodeDefCurrent, mode, la
   return variables
 }
 
+export const getParentsInOrder = ({ parent, survey, includeMultiples = false }) => {
```
ramirobg94, line 80: Method to get the list of the nodes in the survey which has children and are not multiple in order, 
This method is used later to get the list of uuid and the index of each parent

minotogna: what do you mean this method is used later?

ramirobg94: the response of this method is used later*

minotogna: what does this method do?

minotogna (approved):

PR #1299, opened 2020-11-25
Title should be mandatory

### LINKED ISSUE #1129: ExpressionEditor: group attributes by parent entity

### LINKED ISSUE #1286: User registration: Title is mandatory

ramirobg94:
Resolves #1286


server/modules/user/api/userApi.js
```diff
@@ -149,10 +151,16 @@ export const init = app => {
...
-
+    const body = Request.getBody(req)
+    const { user: userToUpdateString } = body
+    let userToUpdate = A.parse(userToUpdateString)
     const validation = await UserValidator.validateUser(userToUpdate)
 
+    userToUpdate = {
+      ...userToUpdate,
+      props: A.stringify(userToUpdate.props),
+    }
```
ramirobg94, line 162: I am worried about that because due to we use formData, we should do this

minotogna: what?

minotogna, line 162: this is unclear. why that conversion?

minotogna (changes requested):

webapp/views/Guest/views/ResetPassword/ResetPassword.js
```diff
@@ -4,6 +4,8 @@ import React from 'react'
 import { useI18n } from '@webapp/store/system'
 import Error from '@webapp/views/Guest/Error'
 
+import DropdownUserTitle from '@webapp/views/App/views/Users/DropdownUserTitle'
```
minotogna, line 7: this is a common component now. move it to common components (under form)

webapp/views/Guest/views/ResetPassword/store/hooks.js
```diff
@@ -17,24 +17,25 @@ export const useResetPassword = () => {
...
-  const { name, password, passwordConfirm } = state.user
+  const { name, password, passwordConfirm, props } = state.user
 
   const { data: { user } = {}, dispatch: getResetPasswordUser } = useAsyncGetRequest(`/auth/reset-password/${uuid}`)
 
   const {
     data: { result: resetComplete = false } = {},
     dispatch: dispatchPostResetPassword,
-  } = useAsyncPutRequest(`/auth/reset-password/${uuid}`, { name, password })
+  } = useAsyncPutRequest(`/auth/reset-password/${uuid}`, { name, props, password })
```
minotogna, line 27: can't you just pass the title ?

server/modules/user/api/userApi.js
ramirobg94: The repository expects that the props are a string, maybe it is better to move this logic to stringify to the repository.

ramirobg94: same as above, I felt dirty when I did that 👎

minotogna: can u make it the same way we persist props for nodeDefs ?
This code might be old... better we fix it now.
what do you think?

ramirobg94: Perfect! I will take a look into the nodeDef code, yep I am finding code that is not used and outdated.

minotogna (approved):

PR #1327, opened 2020-12-10
Create test.js.yml

ramirobg94:


.github/workflows/test.js.yml
```diff
@@ -0,0 +1,47 @@
...
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Use Node.js 14.15.0
+      uses: actions/setup-node@v1
+      with:
+        node-version: '14.15.0'
+    - name: install browser dependencies
+      if: matrix.os == 'ubuntu-latest'
```
ramirobg94, line 21:
```suggestion
```

.github/workflows/test.js.yml
```diff
@@ -0,0 +1,84 @@
...
+    - run: ln -s dist/server.js .
+    - name: run server
+      run: exec pm2-runtime server.js &
+      env:
+        PGHOST: localhost
+        PGPORT: 5444
+        PGDATABASE: of-arena-test
+        PGUSER: arena
+        PGPASSWORD: arena
+        SESSION_ID_COOKIE_SECRET: my-cookie-secret-key
```
sorja, line 72: Should we use github variables?

ramirobg94: yep I thinkt that should be a better way to set env vars

sorja: You can use .env file or use github secrets/env vars

PR #1390, opened 2021-02-17
add this to validation rule

### LINKED ISSUE #1325: Validation rule and list of attributes: show current attribute first (by its name or as $this in Xpath)
Expression Editor in Validation works fine, and grouping too works fine. The issue is that the current (default) variable should be on the top of the dropdown list, even as single varaiable name even though it belongs to a group that is shown somewhere at the end of the list.
In the following picture the attribute name is "plot_slope" but it cannot be easily selected. 

image

ramirobg94:
resolves #1325 
Captura de pantalla 2021-02-17 a las 8 13 45

### DIFF
--- a/webapp/components/expression/expressionVariables.js
+++ b/webapp/components/expression/expressionVariables.js
@@ -42,6 +42,17 @@ const getSqlVariables = (nodeDef, lang) => {
   }))
 }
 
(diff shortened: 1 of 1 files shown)


minotogna (changes requested):
this must be thought carefully. close it for now

PR #1392, opened 2021-02-17
Feature / expressions: entity improvements

### LINKED ISSUE #1367: Expressions: entity improvements
- [x] Allow accessing entities with array indexes and properties inside of them (e.g. cluster.plot[0].tree[1].dbh)
- [x] Add length property (e.g.cluster.plot[1].tree.length)

SteRiccio:
📌 References
Resolves #1367

🥅 What is the goal?
This pull request allows the access of entities using indexes in the expressions.
E.g. plot[1].tree[2].dbh
Also it allows to count the number of multiple nodes (entities or multiple attributes).
E.g. plot[1].tree.length

🚧 How it has been solved?
To allow this, "member" expressions must be supported (expressions like plot[1] or tree.dbh).
The biggest changes are in these files:
- recordExpressionParser.js: _memberEval method has been introduces to evaluate member expressions
- nodeDefExpressionsValidator.js: the validation of a single nodeDef expression has been moved to the nodeDefExpressionValidator module
- nodeDefExpressionValidator / memberEval.js: here the evaluation of member expressions is performed to verify if the expression is semantically correct (member properties must be native node properties or children definition of the current node).
- nodeDefExpressionValidator / identifierEval.js: almost unchanged, moved from the old file nodeDefExpressionsValidator.js
- nodeDefExpressionNativeProperties.js has been introduces to evaluate "native" node properties, like length
- advancedExpressionEditorPopup.js: the validation of the expression has been deleted and will be done using nodeDefExpressionValidator
- codemirrorArenaExpressionHint.js has been modified to autocomplete member expression elements. The available variables will be determined every time the autocomple popup is opened (e.g. in an expression like plot[1]. the available variables will be only the children definition of the entity plot)

✅ Test Cases
Unit tests have been introduced (in 002recordExpressionParser.test.js) to verify that expressions like plot[index] and plot.length work.

‼️ Disclaimer!
The file codemirrorArenaExpressionHint.js is not very clean, I will refactor it a bit in the next pull request.


core/expressionParser/expression.js
```diff
@@ -17,6 +17,12 @@ export const modes = {
   sql: 'sql',
 }
 
+export const nodeProperties = {
+  length: 'length',
+}
```
minotogna, line 22: these properties must be taken from native javascript types.

SteRiccio: yes, but we should know somehow, what are the "allowed" properties, the ones we support, to have control on the expressions written by the user. I'll move them to a separate file.

core/survey/nodeDefExpressionNativeProperties.js
```diff
@@ -0,0 +1,29 @@
...
+
+const nativeProperties = {
+  length: 'length',
+}
+
+const nativePropertiesEvaluators = {
+  length: ({ node }) => {
+    if (Array.isArray(node)) {
+      return node.length
+    }
```
minotogna, line 13: length exists also in string objects. we should be able to let users use native js properties. is this coming in another pull request ?

test/unit/tests/002recordExpressionParser.test.js
```diff
@@ -99,6 +108,17 @@ describe('RecordExpressionParser Test', () => {
...
+    { q: 'plot.length', r: 3 },
+    // access multiple entities with index
+    { q: 'plot[0].plot_id', r: 1 },
+    { q: 'plot[1].plot_id', r: 2 },
+    { q: 'plot[2].plot_id', r: 3 },
+    { q: 'plot[4].plot_id', r: null },
+    // plot_multiple_number counts
+    { q: 'plot[0].plot_multiple_number.length', r: 2 },
+    { q: 'plot[1].plot_multiple_number.length', r: null },
+    { q: 'plot[2].plot_multiple_number.length', r: 1 },
```
minotogna, line 121: we should also test that 
- 'string'.length exist
- number.length throws an error
- 'string'[idx] returns the character at the index idx

SteRiccio: it will be done in another PR

core/survey/nodeDefExpressionNativeProperties.js
SteRiccio: yes, it will be done in another pull request

minotogna (approved):

PR #1422, opened 2021-03-02
Bug/node def default

### LINKED ISSUE #1405: DataEntry: NodeDefTaxon with default value doesn't allow updating value
Steps to reproduce:

1. create a nodeDefTaxon
2. Set a default value
3. goto form preview
4. delete code
5. The value gets updated with values from backend

Expected:
5. allow user to add new value or clear the default value

Test:
Test case is already in surveyFormPreview. uncomment it

ramirobg94:
## Related Issue

Resolves #1405 

When someone is updating a node and in this updating, the user deletes the value, (the new value is empty), the default value is not going to be set.

In this case, seems that the user is trying to clean the value, maybe if she/he likes to recover the default value we should allow it in another way.

## How has this been tested
tests

- [ ] No
- [x] Yes

## Types of changes

- [ ] Docs change 
- [ ] Refactoring 
- [ ] Dependency upgrade
- [x] Bug fix (non-breaking change which fixes an issue)

## Disclaimer

This affects other Nodes too, at the end is a common problem


server/modules/record/manager/_recordManager/nodeUpdateDependentManager.js
```diff
@@ -114,12 +114,16 @@ export const updateSelfAndDependentsDefaultValues = async (survey, record, node,
...
-
+      // 4. persist updated node value if changed and new value is not empty, and return updated node
       if (R.equals(oldValue, exprValue)) {
         return {}
       }
 
+      // if value is empty and node is Taxon leave empty, the user likes to clean the value
+      if (R.isEmpty(oldValue) && NodeDef.isTaxon(nodeDef)) {
+        return {}
+      }
```
minotogna, line 125: I just realised this happens to date, time, taxon, code, boolean, decimal, integer..
we should find a more consistent way to fix this 

Uploading Screen Recording 2021-03-03 at 05.42.32.mov…

minotogna (changes requested):

server/modules/record/manager/_recordManager/nodeUpdateDependentManager.js
ramirobg94: I had the solution for that on the prev commit, but I reduced the scope only for the case of the bug, 

I will make changes to avoid default values on edit.

ramirobg94: the idea with this new line was add some prop to the NodeDEfs like cleanable 

If (R.isEmpty(oldValue) && NodeDef.isCleanable(nodeDef)) 

and allow overwriting with default values on value clean only if the NodeDef is not cleanable

minotogna: I think it's simply that the default value should not be applied on the node users are currently editing.
the concept of cleanable doesn't sound very robust

ramirobg94: okey i will proceed in that way, avoid default on update

webapp/components/survey/SurveyForm/nodeDefs/components/types/nodeDefBoolean.js
```diff
@@ -26,7 +26,7 @@ const NodeDefBoolean = (props) => {
         className="btn btn-s btn-transparent flex-center"
         data-value={value}
         aria-disabled={disabled || !canEditRecord || readOnly}
-        onClick={() => updateNode(nodeDef, node, checked ? null : value)}
+        onClick={() => (checked ? null : updateNode(nodeDef, node, value))}
```
minotogna, line 29: does this prevent resetting a boolean node?

minotogna (changes requested):
we should allow resetting Boolean value

webapp/components/survey/SurveyForm/nodeDefs/components/types/nodeDefBoolean.js
ramirobg94: I changed the test to avoid the reset

minotogna (approved):

PR #1429, opened 2021-03-03
Expressions: allow use of composite attribute members

### LINKED ISSUE #1128: Expression editor: Using individual elements of attribute type taxon, coordinate, date, time
Taxon attribute is actually an array and now it cannot perform correctly in expressions, as in this: 
image

So this attribute 'my_species' should refer to the correct column in taxon table, as follows:
my_species@code
my_species@scientific_name
my_species@family
my_species@genus
(my_species@vernacular_name ??)

So above Expression Editor should look as follows:
my_species@scientific_name == "Pinus sylvestris"

 Please see also this:
http://www.openforis.org/support/questions/633/selecting-individual-elements-of-attribute-type-taxon 

***************************************
Similarly, coordinate data requires similar methods for its elements:
@x
@y
@srs

For example: we can have a validation rule that 
plot_measured_location/@y<60

SteRiccio:
## Related Issue

Resolves #1128

## Brief description of the changes proposed an/or how the issue(s) has(have) been solved.

The changes are in the way an "identifier" is evaluated: if it is a composite attribute value member, than the corresponding value is returned.
I've splitted the valuePropKeys into several objects, depending on the type of attribute (and the use of that constant has been updated accordingly).

## How has this been tested

Unit tests for expression validator and evaluator have been added.

- [ ] No
- [x] Yes

## Types of changes

- [ ] Docs change 
- [ ] Refactoring 
- [ ] Dependency upgrade
- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Disclaimer

This pull request changes the way in which the value of a "composite" attribute is returned by the expression (now in this case the value object is returned, instead of a string representation of it).
Accessing members of Date and Time attributes will be allowed in the next pull request.

### DIFF
--- a/test/unit/tests/002recordExpressionParser.test.js
+++ b/test/unit/tests/002recordExpressionParser.test.js
@@ -133,6 +133,12 @@ describe('RecordExpressionParser Test', () => {
     { q: 'Date.parse(Date()) <= Date.now()', r: true },
     { q: 'Number(remarks)', r: 0 },
     { q: 'String(cluster_id)', r: '12' },
+    // composite attribute members
+    { q: 'cluster_location.x', r: 41.883012 },
+    { q: 'cluster_location.y', r: 12.489056 },
+    { q: 'cluster_location.srs', r: 'EPSG:4326' },
+    { q: 'plot[0].tree[0].tree_species.code', r: 'ACA' },
+    { q: 'plot[0].tree[0].tree_species.scientificName', r: 'Acacia sp.' },
   ]
 
   NodeDefExpressionUtils.testRecordExpressions({ surveyFn: () => survey, recordFn: () => record, queries })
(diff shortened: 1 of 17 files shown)


test/unit/tests/007nodeDefExpressionValidator.test.js
```diff
@@ -73,6 +73,12 @@ describe('NodeDefExpressionValidator Test', () => {
     { q: 'Number(remarks)', r: true },
     { q: 'String(cluster_id)', r: true },
     { q: 'Strings(cluster_id)', r: false },
+    // composite attributes
```
minotogna, line 76: can we add a test case with code ?

test/unit/tests/002recordExpressionParser.test.js
```diff
@@ -133,6 +133,12 @@ describe('RecordExpressionParser Test', () => {
     { q: 'Date.parse(Date()) <= Date.now()', r: true },
     { q: 'Number(remarks)', r: 0 },
     { q: 'String(cluster_id)', r: '12' },
+    // composite attribute members
```
minotogna, line 136: can we have a test case with code?

minotogna (commented):
are composite attribute members visualised in the expression editor?

minotogna (approved):

ramirobg94 (approved):

PR #1432, opened 2021-03-04
Feature/expressions date time members

### LINKED ISSUE #1426: Expression Editor: allow date and time properties. e.g. date.hour, date.year

SteRiccio:
## Related Issue

Resolves #1426

## Types of changes

- [ ] Docs change
- [ ] Refactoring
- [ ] Dependency upgrade

The main changes are in recordExpressionParser/identifierEval.js and nodeDefExpressionValidator/identifier.js
In order to evaluate the node value properties on date and time I've changed the way an identifier is evaluated inside a member expression, so that the expression of the "object" of the member expression is evaluated returning a node instead of it's value and the property needed from the node can be evaluated in the next step of the evaluation.
E.g. `date_attibute.day` is a member expression, the "object" is date_attribute and the property to look for is "day". If the evaluation of the object returns a value like '2021-03-04' it will be tricky to parse it and extract the required property, instead if it returns a node it's quite straightforward.

## How has this been tested

Unit tests have been added.

- [x] No
- [ ] Yes

## Disclaimer

recordExpressionParser/identifierEval.js starts to be complex and difficult to read... I've tried to split its main function but it remains still a sort of salad... we should sit together and see what can be done to simplify it.


core/record/node.js
```diff
@@ -192,6 +205,17 @@ const _getDatePart = (index) => R.pipe(R.partialRight(getValue, ['--']), R.split
 export const getDateYear = _getDatePart(0)
 export const getDateMonth = _getDatePart(1)
 export const getDateDay = _getDatePart(2)
+
```
minotogna, line 208: shouldnt these return numbers?

SteRiccio: yes, I'll change it and I'll check where they are used in the code.

minotogna (approved):

ramirobg94 (approved):

PR #1438, opened 2021-03-22
Expression functions: categoryItemProp

SteRiccio:
## Types of changes

- [ ] Docs change
- [ ] Refactoring
- [ ] Dependency upgrade

Added categoryItemProp function to custom functions in record expression evaluator.

## How has this been tested

Unit tests have been added,

- [ ] No
- [x] Yes

## Disclaimer

Validation of expressions using this function can be improved (e.g. check that the specified categoryName or propName exist)

### DIFF
--- a/core/expressionParser/helpers/functionImplementations.js
+++ b/core/expressionParser/helpers/functionImplementations.js
@@ -5,6 +5,7 @@ import { functionNames } from './functions'
 
 export const functionImplementations = {
   [functionNames.avg]: A.identity,
+  [functionNames.categoryItemProp]: () => null, // To be implemented by evaluator sub classes
   [functionNames.count]: A.identity,
   [functionNames.includes]: (items, value) => Array.isArray(items) && items.includes(String(value)),
   [functionNames.index]: () => 0,
--- a/core/record/recordExpressionParser/recordEpressionFunctions.js
+++ b/core/record/recordExpressionParser/recordEpressionFunctions.js
@@ -1,8 +1,24 @@
+import * as Survey from '@core/survey/survey'
+import * as Category from '@core/survey/category'
+import * as CategoryItem from '@core/survey/categoryItem'
 import * as Record from '@core/record/record'
 import * as Node from '@core/record/node'
 import * as Expression from '@core/expressionParser/expression'
 
-export const recordExpressionFunctions = ({ record }) => ({
+export const recordExpressionFunctions = ({ survey, record }) => ({
+  [Expression.functionNames.categoryItemProp]: (categoryName, itemPropName, ...codePaths) => {
+    const category = Survey.getCategoryByName(categoryName)(survey)
+    if (!category) return null
+
+    const categoryItem = Survey.getCategoryItemByHierarchicalCodes({
+      categoryUuid: Category.getUuid(category),
+      codePaths,
+    })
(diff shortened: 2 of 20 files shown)


minotogna:
Task linked: [CU-fv4jzk ExpressionFunction: categoryItemProp](https://app.clickup.com/t/fv4jzk)

minotogna (commented):
test is missing

minotogna (approved):

core/expressionParser/helpers/functionImplementations.js
```diff
@@ -5,6 +5,7 @@ import { functionNames } from './functions'
 
 export const functionImplementations = {
   [functionNames.avg]: A.identity,
+  [functionNames.categoryItemProp]: () => null, // To be implemented by evaluator sub classes
```
sorja, line 8: Add here (maybe) TODO:
So it can be found easier

Unless of course, you remember this :)

core/survey/_survey/surveyRefDataIndex.js
```diff
@@ -52,83 +52,82 @@ const keys = {
   taxonIndex: 'taxonIndex',
 }
 
+const categoryItemNullParentUuid = 'null'
```
sorja, line 55: Why is this string `'null'` and not just `null`?

core/record/recordExpressionParser/recordEpressionFunctions.js
```diff
@@ -1,8 +1,26 @@
...
+import * as Survey from '@core/survey/survey'
+import * as Category from '@core/survey/category'
+import * as CategoryItem from '@core/survey/categoryItem'
 import * as Record from '@core/record/record'
 import * as Node from '@core/record/node'
 import * as Expression from '@core/expressionParser/expression'
 
-export const recordExpressionFunctions = ({ record }) => ({
+export const recordExpressionFunctions = ({ survey, record }) => ({
+  [Expression.functionNames.categoryItemProp]: (categoryName, itemPropName, ...codePaths) => {
```
sorja, line 10: would it be better to avoid destructuring the codePaths like `...codePaths` ?

sorja (approved):

core/expressionParser/helpers/functionImplementations.js
SteRiccio: the fact is that the implementation is empty; it can be estended by subclasses.
Anyway, we willl get rid of this implementation of the expression evaluator and use the one in arena-core, it's better!

core/survey/_survey/surveyRefDataIndex.js
SteRiccio: because it's used as a key in the dictionary of category items by parent item uuid

core/record/recordExpressionParser/recordEpressionFunctions.js
SteRiccio: it makes the usage of the categoryItemProp function easier for the users.
E.g. categoryItemProp('list_name', 'prop_name', 1, 2, 3) instead of something like categoryItemProp('list_name', 'prop_name', Array.of(1,2,3))
(we cannot use the literal notation to initialize arrays now, like [1,2,3], it's not supported by our expression evaluator)

ramirobg94 (approved):

PR #1468, opened 2021-04-26
Cu hf7bp3 use arena server db ramirobgav

ramirobg94:
## Description

- #### Added/Removed/Updated something...
   
- #### Added/Removed/Updated something else...
    

## Types of changes

- [ ] Docs change
- [ ] Refactoring
- [ ] Dependency upgrade

- [ ] No
- [ ] Yes


minotogna (changes requested):
to avoid huge changes, cant you just export DB as db from server/db/db?
This is the only change I had in mind

.npmrc
```diff
@@ -1,3 +1,3 @@
 scripts-prepend-node-path=true
 @openforis:registry=https://npm.pkg.github.com
-always-auth = true
+always-auth = true
```
minotogna, line 3: restore new line

minotogna (approved):

PR #1543, opened 2021-06-25
Show node def edit buttons only on mouse over

SteRiccio:
In node def table header, show the node def edit buttons only on mouse over and show a tooltip with the full label/name if the text is too long and overflows the width of the column.
Necessary because when node def names are long it is quite difficult to find a node definition.

### DIFF
--- a/webapp/components/survey/SurveyForm/nodeDefs/components/nodeDefTableCellHeader.js
+++ b/webapp/components/survey/SurveyForm/nodeDefs/components/nodeDefTableCellHeader.js
@@ -1,49 +1,15 @@
 import './nodeDefTableCellHeader.scss'
 
-import React, { useEffect, useRef, useState } from 'react'
+import React from 'react'
 import PropTypes from 'prop-types'
 
 import * as NodeDef from '@core/survey/nodeDef'
 
+import { LabelWithTooltip } from '@webapp/components/form/LabelWithTooltip'
 import { useI18n } from '@webapp/store/system'
 
 import * as NodeDefUiProps from '../nodeDefUIProps'
 import NodeDefIconKey from './NodeDefIconKey'
-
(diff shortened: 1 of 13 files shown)


webapp/components/survey/SurveyForm/nodeDefs/components/nodeDefTableCellHeader.js
```diff
@@ -10,21 +10,55 @@ import { useI18n } from '@webapp/store/system'
 import * as NodeDefUiProps from '../nodeDefUIProps'
 import NodeDefIconKey from './NodeDefIconKey'
 
+const LabelWithTooltip = (props) => {
```
ramirobg94, line 13: What do you think about extract this component?

webapp/components/survey/SurveyForm/nodeDefs/nodeDefSwitch.js
```diff
@@ -30,6 +30,13 @@ class NodeDefSwitch extends React.Component {
     super(props)
 
     this.element = React.createRef()
+
+    this.state = {
```
ramirobg94, line 34: Heheh

ramirobg94 (approved):

webapp/components/survey/SurveyForm/nodeDefs/components/nodeDefTableCellHeader.js
SteRiccio: eheheh I didn't want to do it now but... it can be useful elsewhere... how can we call it?!

ramirobg94: I approved because yes maybe it is not the moment, the name is perfect in my opinion

ramirobg94 (approved):

PR #1547, opened 2021-06-28
Avoid no label chain edit

ramirobg94:
## Description

- #### Added/Removed/Updated something...
   
- #### Added/Removed/Updated something else...
    

## Types of changes

- [ ] Docs change
- [ ] Refactoring
- [ ] Dependency upgrade

- [ ] No
- [ ] Yes


webapp/views/App/views/Analysis/Chain/Chain.js
```diff
@@ -48,6 +52,8 @@ const ChainComponent = () => {
 
   return (
     <div className={classNames('chain', { 'with-cycles': cycleKeys.length > 1 })}>
+      <Prompt when={A.isEmpty(chain.props.labels)} message={i18n.t('chainView.errorNoLabel')} />
```
SteRiccio, line 55: you should add even a ChainValidator to validate the form and show a feedback in the form fields; you can use it to get a "validation" object, that can be used in the Prompt "when" condition, like when={!Validation.isValid(validation)}

SteRiccio (changes requested):

SteRiccio (approved):

PR #1602, opened 2021-07-19
Enhancement / survey language selector

SteRiccio:


webapp/views/App/views/Designer/SurveyHierarchy/Tree/Tree.js
```diff
@@ -292,11 +292,28 @@ export default class Tree {
...
-      .text((d) => NodeDef.getLabelWithType({ nodeDef: d.data, lang: this.lang, type: nodeDefLabelType }))
-    this.nodeDefLabelType = nodeDefLabelType
+      .text((d) => NodeDef.getLabelWithType({ nodeDef: d.data, lang: this.lang, type: this.nodeDefLabelType }))
+  }
+
+  get nodeDefLabelType() {
+    return this._nodeDefLabelType
+  }
+
+  set nodeDefLabelType(nodeDefLabelType) {
```
ramirobg94, line 306: is this name right? or should be setNode..

SteRiccio: uhm, no, it's a setter, and from outside it will be used like tree.nodeDefLabelType = ... 
From the outside it will be similar to set a variable in the class called nodeDefLabelType... but we call _updateLabels() method too.

SteRiccio: anyway, the name of the setter for "lang" was not correct! it was not consistent with the getter...

ramirobg94 (approved):

PR #1619, opened 2021-07-26
Bugfix/invalid dates

### LINKED ISSUE #1618: check error opening records in test survey Flora&Fauna

SteRiccio:
fixes #1618


core/dateUtils.js
```diff
@@ -118,6 +118,9 @@ export const convertDate = ({ dateStr, formatFrom = formats.dateISO, formatTo })
     return null
   }
   const dateParsed = parse(dateStr, formatFrom)
+  if (!dateParsed || Number.isNaN(dateParsed.getTime())) {
+    return null
+  }
```
ramirobg94, line 123: Maybe there is a better way to do this

SteRiccio: yes, there is, using the function isValid from date-fns for example

ramirobg94 (approved):

PR #1635, opened 2021-07-29
added app preloader

### LINKED ISSUE #1634: show loader on application startup (it should be visible when connection is quite slow)

SteRiccio:
resolves #1634 
added very simple pure CSS preloader.
the style is different from the Loader already in the app, it should be nice to find a similar one instead.


web-resources/index.html
```diff
@@ -6,9 +6,87 @@
     <meta http-equiv="Content-Type" content="text/html; charset=utf-8"/>
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
+
+    <style>
+      #preloader-wrapper {
+        width: 100vw;
+        height: 100vh;
+        display: flex;
+        background: wh;
```
ramirobg94, line 15:
```suggestion
```

ramirobg94 (approved):

PR #1646, opened 2021-08-04
Enhancement/survey manager role

SteRiccio:


core/auth/authGroup.js
```diff
@@ -46,6 +46,66 @@ export const groupNames = {
   surveyGuest: 'surveyGuest',
 }
 
+const permissionsByGroupName = {
```
ramirobg94, line 49: I think that we have this list in another part of the app, or in arena-core

SteRiccio: updated, the other part of the code uses this one

ramirobg94 (approved):

PR #1761, opened 2021-09-02
use preferred language in chain

### LINKED ISSUE #1755: Chains, selected language: Label and Description texts need to change according to the chosen language
in multilingual survey, in CHAINS section, Label and Description need to change according to the selected language. This is also the case while editing/viewing the result variables.

ramirobg94:
resolves #1755


SteRiccio:
it doesn't solve the problem reported by Lauri!
Also, with this change there won't be the possibility to give the survey default language if not specified in the prefs (dependency loop between UserState and SurveyState)

PR #1814, opened 2021-09-28
Chain changes

### LINKED ISSUE #1765: Changes to UI, RStudio project and logic
The CHAINS allows now to apply a static order in the calculation chain, I mean that it must follow the order of entities in the hierarchy. So currently, only the order of result variables inside the same entity can be changed. However, this is too static system and it will cause problems: there are cases where the categorical result of the parent (e.g.'plot') depends on the computed property of its child (e.g. 'tree').

For example: In order to detect a categorical variable plot$plot_land_category, we need to compute first the sum of trees' basal areas ('tree$tree_basal_area') in the plot (see the the 2nd image below).
I suggest the following changes:

1) while adding a new result variable, the system asks for the entity name (dropdown list). In order to be understandable, this dropdown list needs to repeat the full path of each listed entity. For example
 cluster
 plot
 tree
 stump
 ..

NOTE: Only the multiple entry entities are shown in this dropdown list!

The R script comes visible only after selecting the entity.

2) The list of result variables is common for all entities, and it contains a column for entity name (1st column). As follows:
image

The order of result variables can be changed freely (drag&drop) similarly as now, and the order of computing the result variables should be same in RStudio as in this screen.
image

3) In RStudio project, all variables go into the same folder: /user
(so no distinction between the entities)
image

4) Surveys with existing calculation chains: can we keep them and show them also in the new table? It is not a good idea to rewrite any scripts again.

ramirobg94:
resolves #1765

### DIFF
--- a/core/survey/_survey/surveyNodeDefs.js
+++ b/core/survey/_survey/surveyNodeDefs.js
@@ -111,7 +111,10 @@ export const getNodeDefKeys = (nodeDef) =>
     R.filter((n) => NodeDef.isKey(n) && !NodeDef.isDeleted(n))
   )
 
-export const getNodeDefRootKeys = (survey) => getNodeDefKeys(getNodeDefRoot(survey))(survey)
+export const getNodeDefRootKeys = (survey) => {
+  const root = getNodeDefRoot(survey)
+  return root ? getNodeDefKeys(root)(survey) : []
+}
 
 export const isNodeDefRootKey = (nodeDef) => (survey) =>
   NodeDef.isKey(nodeDef) && NodeDef.isRoot(getNodeDefParent(nodeDef)(survey))
--- a/webapp/components/form/DateInput/ReactDateInput/component.js
+++ b/webapp/components/form/DateInput/ReactDateInput/component.js
@@ -145,25 +145,27 @@ const DateInput = React.forwardRef(
 
     const handleShouldFocusCalendar = (e) => {
       if (!withCalendar) return
-      if (e.target === elRefs[2].current && e.key === KeyboardKeys.Tab) {
-        e.stopPropagation()
-        e.preventDefault()
-        calendarContainer.current.getElementsByTagName('button')[0].focus()
-      }
-
-      if (isCalendarOpen && e.key === KeyboardKeys.Enter) {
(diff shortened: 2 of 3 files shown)


webapp/views/App/views/Analysis/Chain/AnalysisNodeDefs/AnalysisNodeDef/AnalysisNodeDef.js
```diff
@@ -49,6 +49,7 @@ const AnalysisNodeDef = ({ nodeDefUuid }) => {
           <span className="icon icon-14px icon-menu" />
         </button>
       </div>
+      <div>{NodeDef.getName(Survey.getNodeDefByUuid(NodeDef.getParentUuid(nodeDef))(survey))}</div>
```
SteRiccio, line 52: you could use Survey.getNodeDefParent(nodeDef)(survey)

core/survey/survey.js
```diff
@@ -227,3 +229,23 @@ export const {
...
+        NodeDef.isEntity(nodeDef) &&
+        !NodeDef.isSingleEntity(nodeDef) &&
+        getAnalysisNodeDefs({ entity: nodeDef, chain })(survey).length > 0
+      ) {
+        entities.push(nodeDef)
+      }
+    })
+
+    return entities
+  }
```
SteRiccio, line 251: can this go to SurveyAnalysis and export it similarly to getAnalysisNodeDefs?

core/survey/survey.js
```diff
@@ -227,3 +229,23 @@ export const {
...
+  ({ chain }) =>
+  (survey) => {
+    const { root } = getHierarchy()(survey)
+
+    const entities = []
+    entities.push(root)
+    traverseHierarchyItemSync(root, (nodeDef) => {
+      if (
+        NodeDef.isEntity(nodeDef) &&
+        !NodeDef.isSingleEntity(nodeDef) &&
```
SteRiccio, line 243: you can use NodeDef.isMultipleEntity

SteRiccio (changes requested):

SteRiccio (approved):

PR #1907, opened 2021-11-08
visualizer-chart

ramirobg94:
To test this PR a survey with Cluster-plot-tree is needed because some fields are "hardcoded" into the code, 

The panelRight is still WIP and it is going to be the place where we can put the Chart configurator. 

Captura de pantalla 2021-11-08 a las 14 21 27
Captura de pantalla 2021-11-08 a las 14 21 42

Captura de pantalla 2021-11-08 a las 14 21 54


webapp/components/DataQuery/store/useDataQuery.js
```diff
@@ -47,6 +48,13 @@ export const useDataQuery = ({ query }) => {
     Actions.fetch({ offset, limit, query, includesCount: true })
   }, [filter])
 
+  useOnUpdate(() => {
+    const defaultValues = defaults[Query.getDisplayType(query)]
+    setOffset(defaultValues.offset)
+    setLimit(defaultValues.limit)
+    Actions.fetch({ offset, limit, query, includesCount: true })
+  }, [mode])
```
ramirobg94, line 56: maybe not needed

webapp/components/DataQuery/Visualizer/Chart/Chart.js
```diff
@@ -0,0 +1,189 @@
...
+      {chartImage && (
+        <img
+          src={`data:image/svg+xml;base64,${btoa(unescape(encodeURIComponent(chartImage)))}`}
+          alt=""
+          width="100%"
+          height="100%"
+        />
+      )}
+      {showChartEditor && (
+        <PanelRight width="30vw" onClose={() => setShowChartEditor(false)} header={i18n.t('appModules.categories')}>
```
ramirobg94, line 145: this is going to be a panel with 3 modes:

basic -> chose between templates and allow the user to modify some basic fields:
f.e: in arc chart choose the radius, the nodeDef to categorize and if there are some groups the nodeDef of the group
custom -> [ customize the charts based on different properties -> the most complex to implement ]
the most complex to implement ]
advanced -> define the spec using a JSON editor


Now we have a panel on the left to choose the entity and the columns, but now the only thing needed is the entity so maybe we should replace that.

Also maybe is a good idea to try to do this panelRight agnostic of "nodeDefs" I mean give some way to get the variables from outside.

webapp/components/DataQuery/Visualizer/Chart/Chart.js
```diff
@@ -0,0 +1,189 @@
...
+      field: 'cluster_accessibility_label',
+      type: 'nominal',
+      impute: {
+        value: 'NULL',
+      },
+    },*/
+  },
+}
+
+const updateCategoyOnDonnutChart =
```
ramirobg94, line 91: this function should be move to the chart editor, explained into the panel right comment

webapp/components/DataQuery/Visualizer/Chart/Chart.js
```diff
@@ -0,0 +1,189 @@
...
+ TODO, extract this spec configuration to a pannelRight and manange a Chart class
+
+ three modes, basic [ choose between templates and then allow the user to modify some basic fields ]
+ custom [ customize the charts based on different properties -> the most complex to implement ]
+ advance -> define the spec using a JSON editor
+
+ Maybe the only thing that the user needs to chose on the left pannel is the last entity
+*/
+
+const defaultSpec = {
```
ramirobg94, line 23: example of donnut

server/modules/surveyRdb/api/surveyRdbApi.js
```diff
@@ -34,6 +37,32 @@ export const init = (app) => {
     }
   })
 
+  app.post('/surveyRdb/:surveyId/:nodeDefUuidTable/chart', requireRecordListViewPermission, async (req, res, next) => {
```
ramirobg94, line 40: maybe we should extract this to other service to encapsulate Vega

server/modules/surveyRdb/api/surveyRdbApi.js
```diff
@@ -9,6 +9,9 @@ import * as SurveyRdbService from '../service/surveyRdbService'
 
 import { requireRecordListViewPermission, requireSurveyRdbRefreshPermission } from '../../auth/authApiMiddleware'
 
+const vega = require('vega')
+const lite = require('vega-lite')
```
SteRiccio, line 13: maybe it's better to call it vegaLite

server/modules/surveyRdb/api/surveyRdbApi.js
SteRiccio: do you mean the apis? yes, it is something separate from rdb

server/modules/surveyRdb/api/surveyRdbApi.js
```diff
@@ -34,6 +37,32 @@ export const init = (app) => {
...
+        values: rows
+      }}
+
+
+      // TODO extract to Chart Service to encapsulate Vega inside
+      let vegaspec = lite.compile(spec).spec
+      const view = new vega.View(vega.parse(vegaspec), { renderer: 'none' })
+      const svg = await view.toSVG()
+
+      res.json({ svg })
```
SteRiccio, line 60: ah ok, I see it, I would move both services and api to another namespace, something like dataCharts, or "reporting"

PR #1914, opened 2021-11-17
Enhancement/users list last login

SteRiccio:


server/modules/user/repository/userRepository.js
```diff
@@ -75,23 +75,45 @@ export const countUsersBySurveyId = async (surveyId, countSystemAdmins = false,
...
+  const orderBy = orderByFieldBySortBy[sortBy] || 'email'
+  const orderByDirection = sortOrder && sortOrder.toUpperCase() === 'ASC' ? 'ASC' : 'DESC'
+
+  return `
+    WITH us AS (
+      SELECT DISTINCT ON (us.sess #>> '{passport,user}')
+        (us.sess #>> '{passport,user}')::uuid AS user_uuid,
+        (us.expire - interval '30 days') AS last_login_time
+      FROM user_sessions us
+      WHERE us.sess #>> '{passport,user}' IS NOT NULL
```
ramirobg94, line 95: Maybe session is better than `sess`?

server/modules/user/repository/userRepository.js
```diff
@@ -75,23 +75,45 @@ export const countUsersBySurveyId = async (surveyId, countSystemAdmins = false,
...
-export const fetchUsers = async ({ offset = 0, limit = null }, client = db) =>
-  client.map(
-    `
-    SELECT ${selectFieldsCommaSep}
+const _usersSelectQuery = ({ selectFields, sortBy = 'email', sortOrder = 'ASC' }) => {
+  // check sort by parameters
+  const orderByFieldBySortBy = {
+    email: 'email',
+    name: 'name',
+    last_login_time: 'status, last_login_time', // 'status' is used to group users that have accepted invitation, otherwise they have never logged in
```
ramirobg94, line 83: should we use the keys, i think that are repeated between files

server/modules/user/repository/userRepository.js
SteRiccio: eheheh user_sessions is the table used by node-connect-pg-simple... we cannot rename the columns I think.

server/modules/user/repository/userRepository.js
SteRiccio: uhm, these are not the keys but the column names... and in some cases they are the same. Should we introduce an object with the column names?

ramirobg94: yep Let's keep at it is!

ramirobg94 (approved):

PR #1915, opened 2021-11-17
moved NodeDependentsManager functions to RecordNodeDependentsUpdater module

SteRiccio:


core/record/_record/recordNodeDependentsUpdater.js
```diff
@@ -3,16 +3,13 @@ import * as R from 'ramda'
 import * as A from '@core/arena'
 import * as Survey from '@core/survey/survey'
 import * as NodeDef from '@core/survey/nodeDef'
-import * as Record from '@core/record/record'
 import * as Node from '@core/record/node'
 import * as RecordExpressionParser from '@core/record/recordExpressionParser'
 import * as RecordExpressionValueConverter from '@core/record/recordExpressionValueConverter'
+import * as RecordReader from './recordReader'
+import * as RecordUpdater from './recordUpdater'
```
ramirobg94, line 10: should we export from core/record?
should we export as Record.Reader ? or RecordReader?

SteRiccio: uhm, we cannot import Record from this module because we will have a dependency loop... this are like internal modules of Record that should work with other internal modules I think.

ramirobg94: To solve dependency loops usually extraction is the solution

ramirobg94 (approved):

PR #1918, opened 2021-11-19
Node def coordinate: show coordinate on map

SteRiccio:


core/record/node.js
```diff
@@ -199,8 +200,14 @@ export const isValueBlank = (node) => {
...
 // Coordinate
-export const getCoordinateX = _getValuePropRaw(valuePropsCoordinate.x)
-export const getCoordinateY = _getValuePropRaw(valuePropsCoordinate.y)
+export const getCoordinateX = (node) => {
+  const x = _getValuePropRaw(valuePropsCoordinate.x)(node)
+  return R.isNil(x) ? null : Number(x)
+}
+export const getCoordinateY = (node) => {
+  const y = _getValuePropRaw(valuePropsCoordinate.y)(node)
+  return R.isNil(y) || R.isEmpty(y) ? null : Number(y)
```
ramirobg94, line 209: why these validation are different?

ramirobg94: isNill is contained in isEmpty?

SteRiccio: good catch! the check must be the same; R.isEmpty does this actually: 
` return x != null && equals(x, empty(x));`
So it returns false if the specified value (x) is null.

ramirobg94 (approved):

PR #1919, opened 2021-11-19
Base unit dont show base unit in list

ramirobg94:


core/i18n/resources/en.js
```diff
@@ -122,6 +123,7 @@ Do you want to proceed?`,
     save: 'Save',
     saveAndBack: 'Save & Back',
     saved: 'Saved!',
+    see: 'See',
```
SteRiccio, line 126: isn't it better to use "Show" instead of See?

SteRiccio (changes requested):

SteRiccio (approved):

PR #1924, opened 2021-11-24
Enhancement/map view

SteRiccio:
in this pr only the sampling point data at the first level are shown in the map.
In the next prs we can add other levels and record data.


server/modules/expression/api/expressionApi.js
```diff
@@ -10,20 +10,22 @@ import * as Request from '@server/utils/request'
...
+  (item) =>
+    item
+      ? type === NodeDef.nodeDefType.code
+        ? {
+            key: CategoryItem.getCode(item),
+            label: CategoryItem.getLabel(lang)(item),
+          }
+        : {
+            key: Taxon.getCode(item),
+            label: Taxon.getScientificName(item),
```
ramirobg94, line 24: maybe is better to do the if inside of the label

server/modules/expression/api/expressionApi.js
```diff
@@ -10,20 +10,22 @@ import * as Request from '@server/utils/request'
...
+          }
+      : null
```
ramirobg94, line 26:
```suggestion
      ? {
            key: CategoryItem.getCode(item),
            label: type === NodeDef.nodeDefType.code ? CategoryItem.getLabel(lang)(item) : Taxon.getScientificName(item)
          } : null
```

ramirobg94 (approved):

PR #1963, opened 2021-12-06
Chain area based stimate

ramirobg94:


core/survey/_survey/surveyAnalysis.js
```diff
@@ -7,7 +7,7 @@ import { getHierarchy, traverseHierarchyItemSync } from './surveyNodeDefs'
 
 // ====== READ
 export const getAnalysisNodeDefs =
-  ({ chain, entity, entityDefUuid, showSamplingNodeDefs = true, hideSamplingNodeDefsWithoutSibilings = true }) =>
+  ({ chain, entity, entityDefUuid, showSamplingNodeDefs = true, hideSamplingNodeDefsWithoutSibilings = true, hideAreaBasedStimate = true }) =>
   (survey) => {
     let nodeDefs = SurveyNodeDefs.getNodeDefsArray(survey).filter(NodeDef.isAnalysis)
```
SteRiccio, line 12: ideally I would prefer to have one "filter" and several "if" conditions inside the filtering code, to avoid iterating several times the list of node defs (that can be huge, around 300 node defs for example)

SteRiccio (approved):

PR #2115, opened 2022-02-24
Enhancement / category items virtualization

### LINKED ISSUE #2114: Category editor: opening a sampling point data category takes too long

SteRiccio:
fixes #2114 
implemented with a custom component (VirtualizedList), quite simple, but would it be better to use react-virtualized instead?!


ramirobg94:
Maybe is better react-virtualized yes

SteRiccio:
> Maybe is better react-virtualized yes

I've tried to use react-virtualized, but in our case, with the category items that can be "opened" while editing them it doesn't work great, we would apply some workarounds to make it recalculate the height of an item already rendered, and in the end the performance is not great!

PR #2128, opened 2022-03-22
CSV data import: import new records

SteRiccio:


core/record/_record/recordReader.js
```diff
@@ -107,6 +107,38 @@ export const visitDescendantsAndSelf = (node, visitor) => (record) => {
...
+      if (NodeDef.isSingleEntity(nodeDefDescendant)) {
+        return getNodeChildByDefUuid(parentNodeCurrent, descendantDefUuid)(record)
+      } else {
+        throw new Error(
+          `the target node ${NodeDef.getName(nodeDef)} is inside a multiple entity: ${NodeDef.getName(
+            nodeDefDescendant
+          )}`
+        )
+      }
+    }, parentNode)
```
ramirobg94, line 139: can be simple to understand if you move the reduce above of the function return

server/modules/record/manager/_recordManager/recordUpdateManager.js
```diff
@@ -206,7 +202,7 @@ const _updateNodeAndValidateRecordUniqueness = async (
     const { record: recordUpdated1, nodes: nodesUpdated } = await nodesUpdateFn(user, survey, record, node, t)
     let recordUpdated = recordUpdated1
 
-    const { record: recordUpdated2, updatedNodesAndDependents } = await _onNodesUpdate(
+    const { record: recordUpdated2, nodes: updatedNodesAndDependents } = await _onNodesUpdate(
```
ramirobg94, line 205: recordUpdated2 this two and the 1 make me feel weird

webapp/views/App/views/Data/DataImport/DataImportCsvView.js
```diff
@@ -67,8 +89,23 @@ export const DataImportCsvView = () => {
...
+          }))}
+        />
+      </FormItem>
+
       <FormItem label={i18n.t('dataImportView.importIntoEntity')}>
-        <EntitySelectorTree nodeDefUuidActive={selectedEntityDefUuid} onSelect={onEntitySelect} />
+        <EntitySelectorTree
+          nodeDefUuidActive={selectedEntityDefUuid}
+          onSelect={onEntitySelect}
+          isDisabled={() => insertNewRecords}
```
ramirobg94, line 107: is insertNewRecords a function?

SteRiccio: insertNewRecords is a boolean variable, defined few lines above and depending on the dataImportType parameter, but I will change it

server/modules/record/manager/_recordManager/recordUpdateManager.js
SteRiccio: eheheh, I didn't want to change too much the code... I will change it in the next PR! the result of _onNodesUpdate is actually a RecordUpdateResult object, so I can replace it with that (next PR)

core/record/_record/recordReader.js
SteRiccio: maybe I can use a forEach instead... the reduce itself is not so clear in this case

webapp/views/App/views/Data/DataImport/DataImportCsvView.js
```diff
@@ -67,8 +87,23 @@ export const DataImportCsvView = () => {
...
+          }))}
+        />
+      </FormItem>
+
       <FormItem label={i18n.t('dataImportView.importIntoEntity')}>
-        <EntitySelectorTree nodeDefUuidActive={selectedEntityDefUuid} onSelect={onEntitySelect} />
+        <EntitySelectorTree
+          nodeDefUuidActive={selectedEntityDefUuid}
+          onSelect={onEntitySelect}
+          isDisabled={() => dataImportType === importTypes.insertNewRecords}
```
ramirobg94, line 105: ??

SteRiccio: isDisabled should be renamed into isItemDisabled or isNodeDefDisabled, because it's a function that takes a node def and tells if the corresponding item in the tree is disabled...

ramirobg94: but should be a function or just a variable?

SteRiccio: a function, then there should be a "disabled" property to disable the selection at all

PR #2208, opened 2022-05-06
Dependencies / arena server

SteRiccio:


server/modules/auth/api/authApi.js
```diff
@@ -64,14 +47,6 @@ export const init = (app) => {
     }
   })
 
-  app.post('/auth/login', async (req, res, next) => {
```
ramirobg94, line 67: Why login is outside ( in arena-server) but /auth/reset-password not?

SteRiccio: all the other ones could be moved to arena-server too. I have to use the one in arena-server otherwise passport will be initialized 2 times!

ramirobg94 (approved):

PR #2242, opened 2022-05-25
use arena-core with nodes creator (WIP)

SteRiccio:


core/record/_record/recordNodesUpdater.js
```diff
@@ -91,7 +54,7 @@ const _addEntityAndKeyValues =
   ({ survey, entityDef, parentNode, keyValuesByDefUuid }) =>
   (record) => {
     const updateResult = new RecordUpdateResult({ record })
-    const updateResultDescendants = _addNodeAndDescendants({
+    const updateResultDescendants = CoreRecordNodesUpdater.addNodeAndDescendants({
```
ramirobg94, line 57: when do u have to use when you create the rootNode and the new nodes into th multiple entities?

SteRiccio: in theory this one should be used by both, when you create the root entity and when you create an entity, a node that corresponds to a multiple entity definition.

ramirobg94 (approved):

PR #2325, opened 2022-06-24
Feat/basic components

persas:


webapp/views/App/views/Data/Charts/components/Panel/components/BlocksBuilder/BlocksBuilder.js
```diff
@@ -19,7 +26,16 @@ const BlocksBuilder = ({ dimensions, spec, onUpdateSpec }) => {
           {chartsConfig[configKey].selector.title}
         </button>
       ))}
-      <p>{type}</p>
+      {type &&
+        chartsConfig[type].builderBlocks.order.map((blockKey) =>
```
SteRiccio, line 30: can you extract chartsConfig[type].builderBlocks to a variabla, just to make it a bit more readable?

SteRiccio:
ok for me in general, I would clean a bit the code of BlockBuilder because it starts to be a bit complex, but the rest is ok.

ramirobg94 (approved):

PR #2341, opened 2022-06-30
ability to change panel size

persas:


webapp/views/App/views/Data/Charts/components/Panel/components/RawChartBuilder/RawChartBuilder.js
```diff
@@ -127,7 +127,7 @@ const RawChartBuilder = ({ spec, onUpdateSpec, dimensions }) => {
   }, [helpOpened])
 
   return (
-    <div className="raw-chart-builder">
+    <div className={`raw-chart-builder ${visible ? 'visiable' : ''}`}>
```
SteRiccio, line 130: there is a typo here: "visiable". Why don't you use classNames for this? it would become something like: 
className={classNames( 'raw-chart-builder', { visible } ) }

ramirobg94 (approved):

PR #2349, opened 2022-07-07
Feat/select

persas:


webapp/views/App/views/Data/Charts/state/hooks/useChartConfig.js
```diff
@@ -9,9 +9,28 @@ const defaultConfig = {
 const getItemsByPath = (config) =>
   (config?.items || []).reduce((acc, item) => Object.assign({}, acc, { [item.blockPath]: item }), {})
 
+const addItemByPath = (blockPath, item) => (obj) => {
+  let _value = obj[blockPath]?.value || []
+  _value.push(item)
+  obj[blockPath] = { ...(obj[blockPath] || { blockPath }), value: _value }
```
SteRiccio, line 15: this way of assigning the "value" property in that object is repeated many times.. can it be extracted to a function?

PR #2359, opened 2022-07-27
Bug/publish

ramirobg94:


server/modules/surveyRdb/repository/dataView/create.js
```diff
@@ -91,6 +113,8 @@ export const createDataView = async ({ survey, nodeDef }, client) => {
     }
   }
 
+  groupByColumns.push(..._getGroupByColumns(viewDataNodeDef))
```
SteRiccio, line 116: move also the code above into getGroupByColumns function

SteRiccio (changes requested):

.npmrc
```diff
@@ -1,3 +0,0 @@
-scripts-prepend-node-path=true
-@openforis:registry=https://npm.pkg.github.com
-always-auth=true
```
SteRiccio, line 3: don't delete this file, it doesn't contain any personal information and allows every user that dowloads the source code to build it

SteRiccio (changes requested):

server/modules/surveyRdb/repository/dataView/create.js
```diff
@@ -38,6 +38,28 @@ const _getSelectFieldKeys = (viewDataNodeDef) => {
...
+  //  get parent keys names to groupby
+  if (viewDataParent) {
+    const keys = Survey.getNodeDefKeys(viewDataParent.nodeDef)(viewDataParent.survey)
+      .map((nodeDef) => {
+        const columnNodeDef = new ColumnNodeDef(viewDataParent, nodeDef)
+        return `${viewDataParent.alias}.${columnNodeDef.name}`
+      })
+      .flat()
+
+    groupByColumns.push(...(keys || []))
```
SteRiccio, line 58: keys is always not null, it's the result of the flat function (you can even use the flatMap function in this case)

SteRiccio, line 58: I think it's enough to add this to the group by generated before, without adding all the select fields that are not aliased, like in line 46

SteRiccio (approved):

PR #2363, opened 2022-07-29
Mobile upload

ramirobg94:


server/modules/mobile/api/mobileApi.js
```diff
@@ -5,7 +5,7 @@ import * as JobUtils from '@server/job/jobUtils'
 import * as ArenaMobileImportService from '../service/arenaMobileImportService'
 
 export const init = (app) => {
-  // ====== UPDATE - calculated entity data
+  // ====== READ - all survey day
```
SteRiccio, line 8: all survey day?!

SteRiccio:
since you are working on this part, I don't like the idea of having a "mobile" api at all... these functions should go under the recordApi module (record update) and surveyApi (survey download).

server/modules/mobile/api/mobileApi.js
ramirobg94: typo

ramirobg94:
> since you are working on this part, I don't like the idea of having a "mobile" api at all... these functions should go under the recordApi module (record update) and surveyApi (survey download).

I think that having an entry point for mobile is useful because you never know if you need to do some adaptation. Now the endpoint points to the common code. I agree that if we follow REST API we should use endpoints for each resource but if we analyze the code that we have now we have endpoints for R-studio and collect-import.

SteRiccio (approved):

PR #2364, opened 2022-08-02
Refactor/dropdown

SteRiccio:
main changes in Dropdown.js component.
Most of the changes are related to the change of the "items" prop (to be consistent with react-select naming convention).
Next steps:
- split Dropdown and Autocomplete component
- use react-select in InputChips or allow multiple selection to Dropdown
- use the Dropdown component instead of react-select and `<select>` elements


core/app/languages.js
```diff
@@ -598,9 +598,9 @@ export const getLanguageISO639part2Label = (lang, translationLang = 'en') => {
 
 export const languages = R.pipe(
   R.keys,
-  R.map((lang) => ({ key: lang, value: getLanguageLabel(lang) }))
+  R.map((lang) => ({ value: lang, label: getLanguageLabel(lang) }))
```
ramirobg94, line 601: you can retain the key or id, maybe it is useful for some purpose

core/arena/isNull.js
```diff
@@ -7,4 +7,4 @@ import { _curry1 } from './internal/_curry1'
  *
  * @returns {boolean} - True if the input value is `null` or `undefined`, false otherwise.
  */
-export const isNull = _curry1((value) => value == null)
+export const isNull = _curry1((value) => value === null || value === undefined)
```
ramirobg94, line 10: you already have this in arena-core

webapp/components/form/Dropdown/Dropdown.js
```diff
@@ -1,143 +1,96 @@
...
-  }, [itemLabelFunction])
-
-  const showDialog = State.getShowDialog(state)
-  const itemsDialog = State.getItemsDialog(state)
-  const inputValue = State.getInputValue(state) || ''
-
-  const searchMinCharsReached = autocompleteMinChars <= 0 || inputValue.trim().length >= autocompleteMinChars
+  const { inputId, loading, menuIsOpen, onChange, onInputChange, openMenuOnClick, options, selectRef, value } =
+    useDropdown({
+      autocompleteMinChars,
```
ramirobg94, line 48: MinCharacters, It took me 3 seconds to guess what was Chars
or minCharactersToAutocomplete which is more descriptive

webapp/components/form/Dropdown/Dropdown.js
```diff
@@ -1,143 +1,96 @@
...
-        ref={inputRef}
+      <ReactSelect
+        className={classNames('dropdown', className)}
+        classNamePrefix="dropdown"
+        components={{ Option: OptionComponent, IndicatorsContainer: IndicatorsContainerComponent }}
+        inputId={inputId}
+        isClearable={clearable && !readOnly}
+        isDisabled={disabled}
+        isLoading={loading}
+        isSearchable={!readOnlyInput && !readOnly}
```
ramirobg94, line 74: do we have any way to put this false if we have an input and we don't like to have that like searchable?

webapp/components/form/Dropdown/Dropdown.js
```diff
@@ -1,143 +1,96 @@
...
-        onChange={async (value) => {
-          await Actions.updateInputValue({ value, state })
-        }}
-        onBlur={async (e) => {
-          await Actions.onBlurInput({ value: e.target.value, state, selection })
-        }}
-        onFocus={async () => Actions.openDialog({ state })}
-        title={title}
+        ref={selectRef}
+        value={value}
```
ramirobg94, line 82: do we like to have the select as controlled or uncontrolled?

webapp/components/form/Dropdown/useDropdown.js
```diff
@@ -0,0 +1,121 @@
...
+  const inputValue = selectRef?.current?.inputRef?.value
+  const searchMinCharsReached = !autocompleteMinChars || autocompleteMinChars <= inputValue?.trim()?.length
+
+  const getOptionLabel = useCallback(
+    (item) => (itemLabel.constructor === String ? A.prop(itemLabel, item) : itemLabel(item)),
+    [itemLabel]
+  )
+
+  const getOptionValue = useCallback(
+    (item) => (itemKey.constructor === String ? A.prop(itemKey, item) : itemKey(item)),
```
ramirobg94, line 29: itemKey or itemValue

webapp/components/form/Dropdown/useDropdown.js
```diff
@@ -0,0 +1,121 @@
...
+        onChangeProp(item)
+      }
+    },
+    [getItemFromOption, onBeforeChange, onChangeProp]
+  )
+
+  const onInputChange = useCallback(
+    (inputValue) => {
+      if (autocompleteMinChars) {
+        if (autocompleteMinChars <= inputValue?.length) {
```
ramirobg94, line 90: unnecessary nested complexity,

if(!) return 
if(X)
else X

webapp/components/form/Dropdown/useDropdown.js
```diff
@@ -0,0 +1,121 @@
...
+
+  const getItemFromOption = useCallback(
+    (option) => (option ? items.find((itm) => getOptionValue(itm) === option.value) : null),
+    [items, getOptionValue]
+  )
+
+  const onChange = useCallback(
+    async (option) => {
+      const item = getItemFromOption(option)
+      if (!onBeforeChange || (await onBeforeChange(item))) {
```
ramirobg94, line 80: onBeforeChange has the name of an event handler, I am not expect a value as response

webapp/components/form/Select/utils.js
```diff
@@ -1,4 +1,4 @@
 import * as A from '@core/arena'
 
 export const adaptSelection = (value) =>
```
ramirobg94, line 3: selection vs value

webapp/components/survey/CategorySelector/CategorySelector.js
```diff
@@ -40,13 +40,19 @@ export const CategorySelector = (props) => {
   const [showCategoriesPanel, setShowCategoriesPanel] = useState(false)
   const [categoryToEdit, setCategoryToEdit] = useState(null)
 
-  const emptyItem = { uuid: null, label: i18n.t('common.notSpecified') }
+  const emptyItem = useMemo(() => ({ uuid: null, label: i18n.t('common.notSpecified') }), [i18n])
```
ramirobg94, line 43: how often does the i18n change when this component is rendered?

webapp/views/App/views/Analysis/Chain/PValueSelector/PValueSelector.js
```diff
@@ -9,7 +9,7 @@ import { FormItem } from '@webapp/components/form/Input'
 import { useI18n } from '@webapp/store/system'
 import { ChainActions, useChain } from '@webapp/store/ui/chain'
```
ramirobg94, line 11: what is P?

core/app/languages.js
SteRiccio: uhm, no, the key is value, we never use a key different from the value... it will be more consistent with react-select naming convention

core/arena/isNull.js
SteRiccio: yes, but I didn't want to change too much in this PR..

webapp/components/form/Dropdown/Dropdown.js
SteRiccio: yes, but maybe it would be even better if all this logic will be moved to another component, Autocomplete for example, that wraps the dropdown adding this autocompletion logic

webapp/components/form/Dropdown/Dropdown.js
SteRiccio: readOnlyInput=true?

webapp/components/form/Dropdown/Dropdown.js
SteRiccio: controlled

webapp/components/form/Dropdown/useDropdown.js
SteRiccio: right

webapp/components/form/Dropdown/useDropdown.js
SteRiccio: yes, I thought the same, but it is the way it was working before: the callback passed returns a value... it could be changed (another PR)

webapp/components/form/Select/utils.js
SteRiccio: yes, another PR!

webapp/components/survey/CategorySelector/CategorySelector.js
SteRiccio: never, it was just to avoid eslint to complain about a missing dependency!

webapp/views/App/views/Analysis/Chain/PValueSelector/PValueSelector.js
SteRiccio: it's the probability value, quite common in analysis

core/app/languages.js
ramirobg94: If we are not going to use a key different from value maybe, but at this point, we are into the core/app/languages an maybe it is better to have 

key/id, value, label

core/arena/isNull.js
ramirobg94: okey

webapp/components/form/Dropdown/Dropdown.js
ramirobg94: yes that could be better, in this case just change this small name makes a big difference

webapp/components/form/Dropdown/Dropdown.js
ramirobg94: readOnlyInput and searchable are not comparable

readOnlyInput:  is always not searchable and not mutable
but
searchable:  is mutable and searchable 
but ! searchable:  could be mutable or not and not searchable 

as you have readOnly maybe readOnlyInput is not needed, true that makes me confused. why readOnly and readOnlyInput? what is the difference?

webapp/components/form/Dropdown/Dropdown.js
ramirobg94: oki

webapp/components/survey/CategorySelector/CategorySelector.js
ramirobg94: It shouldn't cry before,

webapp/views/App/views/Analysis/Chain/PValueSelector/PValueSelector.js
ramirobg94: oki

core/app/languages.js
SteRiccio: and how will you use the key/id then? the value in that case is the language code, that is unique already

webapp/components/form/Dropdown/Dropdown.js
SteRiccio: ok

webapp/components/form/Dropdown/Dropdown.js
SteRiccio: readOnlyInput in this case means that you cannot write anything in the input field, and I think it's basically the same thing of using isSearchable (the opposite of it actually), while readOnly means that the dropdown is disabled, but it won't be rendered with a disabled style...

webapp/components/survey/CategorySelector/CategorySelector.js
SteRiccio: I'll remove it, it's unnecessary

webapp/components/form/Dropdown/Dropdown.js
SteRiccio: ah another thing, readOnlyInput is always not searchable and could be mutable or not, like searchable.
Maybe it would be easier to change directly readOnlyInput into searchable and pass it to the react-select component, in the way we use it it won't make any difference

ramirobg94: readOnlyInput -> is not mutable and then not searchable, 
but !readOnlyInput could be searchable or not 

______________| w  | r | s | ( write, read, search )
readOnlyInput__|   0 1 0   | ( you can't type so !searchable)
!readOnlyInput_.|   1 1 x    | ( you can or can't have the possibility to search)
searchable_____|   1 1 1    |
!searchable____.|   x 1 0   | ( you can or can't have the possibility to edit the field)

for the searchable property:
readOnlyInput = !searchable  => 0 = 0 
but
!readOnlyInput =/= searchable => X ?= 1

To be comparable and to ensure that they are the opposite, It is necessary that the following be true:
A = !B => !A = B 
and this case is not totally true due to the previous point

Therefore maybe it is better to use isSearchable or searchable alone instead of mixing that with readability.

isSearchable={ isSearchable }

PR #2369, opened 2022-08-10
Geospatial module

Ukonhattu:
## Description

- #### Added Sampling Polygons
   Added Sampling polygons to records. These can be rectangles or circles and they have control points inside them. Control points change color on click. Added options about these to the survey info form.

- #### Added side-by-side
    Added a dual-view of planet NICFI data where you can set different periods in RGB or false color format, next to each other and compare them with a movable slider.

- #### Added KML/KMZ/Shapefile
 Added ability to upload KML/KMZ/Shapefile data on top of the map.

## Types of changes

- [ ] Docs change
- [ ] Refactoring
- [ ] Dependency upgrade

## How has this been tested

These have been tested on local machine, manually going trough all options.

Windows 11
Google Chrome 104.0.5112.81 (64-bit)

- [x] No
- [ ] Yes

## UI changes

kuva
kuva
kuva
kuva
kuva


core/survey/_survey/surveyInfo.js
```diff
@@ -71,6 +72,8 @@ export const getLabel = (surveyInfo, lang) => {
   return StringUtils.isBlank(label) ? getName(surveyInfo) : label
 }
 
+export const getSamplingPolygon = ObjectUtils.getProp(keys.samplingPolygon, { value: '{}' })
```
SteRiccio, line 75: should this be by default an empty object or an object with a property "value" like you have defined here?

core/survey/survey.js
```diff
@@ -47,6 +47,7 @@ export const newSurvey = ({
     [SurveyInfo.keys.languages]: languages,
     [SurveyInfo.keys.labels]: label ? { [languages[0]]: label } : {},
     [SurveyInfo.keys.srs]: srs && srs.length > 0 ? srs : [R.omit([Srs.keys.wkt], Srs.latLonSrs)],
+    [SurveyInfo.keys.samplingPolygon]: SurveyDefaults.getSamplingPolygonDefaults(),
```
SteRiccio, line 50: I wouldn't include the sampling polygon object to all the surveys; it will be used only in some cases and we will end up with a lot of surveys having this "dirty" information...

package.json
```diff
@@ -126,8 +127,10 @@
     "i18next": "^21.8.8",
     "is-valid-coordinates": "^1.0.0",
     "jsonwebtoken": "^8.5.1",
+    "jszip": "^3.10.0",
```
SteRiccio, line 130: not another zip library please, we are including already node-stream-zip, wrapped by utils/file/fileZip.js

webapp/components/Map/KmlUploader.js
```diff
@@ -0,0 +1,142 @@
...
+      } else if (selectedFile.name.endsWith('.zip')) {
+        const geo = L.geoJson(
+          { features: [] },
+          {
+            onEachFeature: function popUp(f, l) {
+              var out = []
+              if (f.properties) {
+                for (var key in f.properties) {
+                  out.push(key + ': ' + f.properties[key])
+                }
```
SteRiccio, line 49: it could be more readable with something like this: 
const out = Object.entries(f.properties).map(...

webapp/components/Map/KmlUploader.js
```diff
@@ -0,0 +1,142 @@
...
+        const reader = new FileReader()
+        reader.onload = async (e) => {
+          const text = e.target.result
+          shp(text).then(function (data) {
+            geo.addData(data)
+          })
+        }
+        reader.readAsArrayBuffer(selectedFile)
+      }
+    }
```
SteRiccio, line 65: please split this into different function calls for each file extension

webapp/components/Map/MapBaseLayerPeriodSelector.js
```diff
@@ -50,44 +100,74 @@ export const MapBaseLayerPeriodSelector = () => {
...
+      setState((statePev) => ({ ...statePev, leftChecked: !statePev.leftChecked, procLeft: newProc }))
+    } else {
+      newProc = !rightChecked ? Proc.cir : Proc.rgb
+      setState((statePev) => ({ ...statePev, rightChecked: !statePev.rightChecked, procRight: newProc }))
+    }
+    let layer
+    if (isLeft) {
+      layer = Object.values(map._layers).find((layer) => layer.options.id !== 'side')
+    } else {
+      layer = Object.values(map._layers).find((layer) => layer.options.id === 'side')
```
SteRiccio, line 153: this way of getting the layer is repeated above, can you extract it into a function?

webapp/views/App/views/Data/MapView/CoordinateAttributeDataLayer/CoordinateAttributeMarker.js
```diff
@@ -37,33 +43,48 @@ export const CoordinateAttributeMarker = (props) => {
...
+          ancestorsKeys={ancestorsKeys}
+          onRecordEditClick={onRecordEditClick}
+        />
+      </CircleMarker>
+    </div>
   )
 }
+
+CoordinateAttributeMarker.propTypes = {
+  ancestorsKeys: PropTypes.any,
```
SteRiccio, line 81: the type should be PropTypes.array, so you avoid to define propTypes for ancestorKeys too

SteRiccio:
leaflet-side-by-side can be added as a dependency, instead of embedding its source code. Am I wrong?

webapp/views/App/views/Home/SurveyInfo/SamplingPolygonEditor/SamplingPolygonEditor.js
```diff
@@ -0,0 +1,262 @@
...
+    setSamplingPolygon(JSON.stringify(samplingPolygonObject))
+  }
+  const controlPointOffsetNorthOnChange = (value) => {
+    samplingPolygonObject.controlpoint_offset_north = value
+    setSamplingPolygon(JSON.stringify(samplingPolygonObject))
+  }
+  const constrolPointOffsetEastOnChange = (value) => {
+    samplingPolygonObject.controlpoint_offset_east = value
+    setSamplingPolygon(JSON.stringify(samplingPolygonObject))
+  }
```
SteRiccio, line 249: since the only part that changes is the "key" of the property in the samplingPolygonObject, I would do it in a different way.
I would create an array like this:
const samplingPolygonProperties = [{ key: 'length_latitude´, labelKey: 'lengthLatitude' }, ...]

then to render the items: 
samplingPolygonProperties.map(({key, labelKey}) => {
const inputId = `survey-info-sampling-control-point-${key}`
return <div className="form-item" key={key}>
        <label className="form-label" htmlFor={inputId}>
          {i18n.t(`samplingPolygonOptions.${labelKey}`)}
        </label>
        <Input
          id={inputId}
          value={samplingPolygonObject[key]}
          validation={getFieldValidation(samplingPolygonObject[key]}
          onChange={value => onPropertyChange({key, value})}
          readOnly={readOnly}
        />
      </div>
)

and the onPropertyChangeFunction will be generic, like: 
 const onPropertyChange = ({key, value}) => {
    samplingPolygonObject[key] = value
    setSamplingPolygon(JSON.stringify(samplingPolygonObject))
  }

another question: could the sampliing polygon object be put into the survey properties as an object instead? why is it "stringified"?

core/survey/_survey/surveyInfo.js
Ukonhattu: Correct, forgot to change this from the first version that worked with value

webapp/components/Map/KmlUploader.js
```diff
@@ -0,0 +1,142 @@
...
+    layers.forEach((layer) => layer.setStyle({ fillOpacity: value / 100 }))
+  }
+
+  const rangeOnMouseDown = () => {
+    map.dragging.disable()
+  }
+
+  const rangeOnMouseUp = () => {
+    map.dragging.enable()
+  }
```
ramirobg94, line 115: maybe you can shortcut both functions below

webapp/components/Map/MapBaseLayerPeriodSelector.js
```diff
@@ -32,14 +40,56 @@ export const MapBaseLayerPeriodSelector = () => {
...
+    selectedPeriodValueRight,
+    leftChecked,
+    rightChecked,
+    procLeft,
+    procRight,
+  } = state
+
+  const initSideBySide = () => {
+    if (User.isSystemAdmin(user)) {
+      const layerLeft = Object.values(map._layers).find((layer) => layer.options.id !== 'side')
```
ramirobg94, line 71: why is the id 'side'? maybe layerPosition: 'left'

webapp/components/Map/MapBaseLayerPeriodSelector.js
```diff
@@ -50,44 +100,74 @@ export const MapBaseLayerPeriodSelector = () => {
...
+    (isLeft) => (event) => {
       const periodValue = event.target.value
       // replace url in current layer
-      const layer = Object.values(map._layers).find((layer) => layer._url && layer.id === contextBaseLayer.id)
+      let layer
+      if (isLeft) {
+        layer = Object.values(map._layers).find((layer) => layer.options.id !== 'side')
+      } else {
+        layer = Object.values(map._layers).find((layer) => layer.options.id === 'side')
+      }
```
ramirobg94, line 120: this code is repeated maybe this code should go into the Map

webapp/components/Map/MapOptions.js
```diff
@@ -36,6 +36,24 @@ export const MapOptions = () => {
                 onChange={(value) => onOptionUpdate({ option: 'showMarkersLabels', value })}
               />
             </FormItem>
+            <FormItem label={i18n.t('mapView.options.showSamplingPolygon')}>
+              <Checkbox
+                checked={options.showSamplingPolygon}
+                onChange={(value) => onOptionUpdate({ option: 'showSamplingPolygon', value })}
+              />
```
ramirobg94, line 43: you can use currification to avoid re-renders

const a = (x) => (value) => doSmthg with x and value

<component onE={a(xValue)} />

webapp/components/Map/ShowZoomLevel.js
```diff
@@ -0,0 +1,26 @@
...
+  const user = useUser()
+  const [zoomLevel, setZoomLevel] = useState()
+
+  useEffect(() => {
+    setZoomLevel(map.getZoom())
+  }, [])
+
+  map.on('zoomend', () => {
+    setZoomLevel(map.getZoom())
+  })
```
ramirobg94, line 20: do you need to remove this listener?

webapp/views/App/views/Data/MapView/CoordinateAttributeDataLayer/CoordinateAttributePolygon.js
```diff
@@ -0,0 +1,245 @@
...
+        return nCircles(middle, radius / 2, 4, 45)
+      case 5:
+        return fiveCircles(middle, radius)
+      case 10:
+        return tenCircles(middle, radius)
+      case 12:
+        return twelveCirles(middle, radius)
+      case 21:
+        return twentyOneCircles(middle, radius)
+    }
```
ramirobg94, line 81: I think that this can be simplified  maybe using just this function: nCircles

Ukonhattu: The circle positions are different depending on the number of circles. If I use only one nCircles function then I'd need to have many if..else's there to handle all the different positions

webapp/components/Map/ShowZoomLevel.js
Ukonhattu: What do you mean?

webapp/components/Map/MapOptions.js
Ukonhattu: True, just did like that since it is not a long function so it does not affect performance. Will change.

core/survey/_survey/surveyDefaults.js
```diff
@@ -50,3 +50,18 @@ export const getDefaultAuthGroups = () => [
...
+    "length_longitude": "0",
+    "number_of_points_north": "0",
+    "number_of_points_east": "0",
+    "offset_north": "0",
+    "offset_east": "0",
+    "controlpoint_offset_north": "0",
+    "controlpoint_offset_east": "0",
+    "isCircle": "true",
+    "radius": "0"
+  }`
```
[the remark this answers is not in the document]
Ukonhattu: This is because I used the existing methods to put data from form to the db and they did make it a string. Have changed it now.

webapp/components/Map/KmlUploader.js
Ukonhattu: What do you mean?

webapp/views/App/views/Home/SurveyInfo/SamplingPolygonEditor/SamplingPolygonEditor.js
Ukonhattu: Yeah correct. Made them own functions because I thought they would need to be different and forgot to change it in the end when it was apparent that they are the same. It was stringified since when I put it to the db the methods stringified it and I thought that your bindings always stringify them but found that it is not true and have changed it now.

webapp/components/Map/KmlUploader.js
Ukonhattu: Oh yeah, been doing so much c# that didn't automatically think in a functional way, which I should in modern js

webapp/components/Map/KmlUploader.js
```diff
@@ -0,0 +1,142 @@
...
+      const track = new L.KML(kml)
+      map.addLayer(track)
+      setLayers((old) => [...old, track])
+    })
+  }
+
+  const fileChangeHandler = (event) => {
+    setSelectedFile(event.target.files[0])
+
+    setIsFilePicked(true)
```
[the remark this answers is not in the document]
Ukonhattu: Perhaps. I had some reason for this but honestly can't remember anymore. Will check if it is needed.

webapp/components/Map/ShowZoomLevel.js
Ukonhattu: Oh I think there is a confusion because of the name? Maybe should change that. setZoomLevel sets the number value on the component that is shown on top of a map, to show what is is the current zoom level. Its "set" because that is the standard way to name state hooks. It does not set the zoom of the map object.

webapp/components/Map/MapBaseLayerPeriodSelector.js
Ukonhattu: No reason for this to go to the map when this is especially for this component and is the main thing of this component.

webapp/components/Map/MapOptions.js
Ukonhattu: Also, I just literally followed the convention there already was, the existing code was like that and I just added free more

webapp/components/Map/KmlUploader.js
ramirobg94: use into the component `onSmth={map.dragging.enable}`

webapp/components/Map/MapBaseLayerPeriodSelector.js
ramirobg94: so instead of being named as 'side' maybe is better to use left-layer and right-layer
or maybe if you like to scale the system to have a grid some positional id, like:

'left', 'center', 'right'

'00' '01' '02'
'10' '11' '12'
'20' '21' '22'

webapp/components/Map/ShowZoomLevel.js
ramirobg94: the map.on('event', is the listener, not sure if you need to remove the listener when you unmount the component

webapp/components/Map/KmlUploader.js
```diff
@@ -0,0 +1,142 @@
...
+
+  const [selectedFile, setSelectedFile] = useState()
+  const [isFilePicked, setIsFilePicked] = useState(false)
+  const [layers, setLayers] = useState([])
+  const [opacity, setOpacity] = useState(50)
+
+  useEffect(() => {
+    if (isFilePicked) {
+      if (selectedFile.name.endsWith('.kmz')) {
+        handleKMZ(selectedFile)
```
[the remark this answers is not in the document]
Ukonhattu: Technically it does the job of a file input handler, but true in the context of js this isn't really a handler in the truest sense.

webapp/components/Map/KmlUploader.js
Ukonhattu: onMouseDown={map.dragging.enable} throws an error, does not work

webapp/components/Map/ShowZoomLevel.js
Ukonhattu: Well it needs to live as long as the map lives, so unless there is a map destroyer somewhere, then I doubt

webapp/components/Map/KmlUploader.js
```diff
@@ -0,0 +1,145 @@
...
+      const text = e.target.result
+      addKMLLayers([text])
+    }
+    reader.readAsText(file)
+  }
+
+  const processKMZFile = async (file) => {
+    const kmlList = []
+    let promises = []
+    await jszip.loadAsync(file).then(() => {
```
SteRiccio, line 75: why do you have an "await" and then a call to "then"?
you should be able to do just: 
await jszip.loadAsync(file)
jszip.forEach(...
await Promise.all(promises)
addKMLLayers(kmlList)

webapp/views/App/views/Data/MapView/CoordinateAttributeDataLayer/CoordinateAttributePolygon.js
```diff
@@ -0,0 +1,245 @@
...
+import GeometryUtil from 'leaflet-geometryutil'
+import L from 'leaflet'
+import { useMapContextOptions } from '@webapp/components/Map/MapContext'
+import { useSurveyInfo } from '@webapp/store/survey'
+import _ from 'lodash'
+import PropTypes from 'prop-types'
+
+export const CoordinateAttributePolygon = (props) => {
+  const surveyInfo = useSurveyInfo()
+  const polygonProperties = JSON.parse(surveyInfo.props.samplingPolygon)
```
SteRiccio, line 12: for already existing surveys, surveyInfo.props.samplingPolygon is null, please check that, or use a getter that returns an empty object in that case. Also the JSON parsing shouldn't be needed because surveyPolygon should be already an object

webapp/components/Map/L.KML.js
```diff
@@ -0,0 +1,480 @@
+var L = require('leaflet')
+
+L.KML = L.FeatureGroup.extend({
```
ramirobg94, line 3: could you use https://github.com/windycom/leaflet-kml ?

webapp/components/Map/MapBaseLayerPeriodSelector.js
```diff
@@ -32,14 +40,62 @@ export const MapBaseLayerPeriodSelector = () => {
...
+
+  const getLayer = (isLeft) => {
+    if (isLeft) {
+      return Object.values(map._layers).find((layer) => layer.options.id !== 'right')
+    } else {
+      return Object.values(map._layers).find((layer) => layer.options.id === 'right')
+    }
+  }
+  const initSideBySide = () => {
+    if (User.isSystemAdmin(user)) {
```
ramirobg94, line 77: you can reduce the complexity and the nested levels if you type:

if(!User.isSystemAdmin(user)) return
...

webapp/components/Map/MapBaseLayerPeriodSelector.js
```diff
@@ -50,44 +106,74 @@ export const MapBaseLayerPeriodSelector = () => {
...
+    (isLeft) => (event) => {
       const periodValue = event.target.value
       // replace url in current layer
-      const layer = Object.values(map._layers).find((layer) => layer._url && layer.id === contextBaseLayer.id)
+      let layer
+      if (isLeft) {
+        layer = getLayer(true)
+      } else {
+        layer = getLayer(false)
+      }
```
ramirobg94, line 126: layer = getLayer(isLeft)

package.json
```diff
@@ -129,6 +130,7 @@
     "jsonwebtoken": "^8.5.1",
     "jwk-to-pem": "^2.0.4",
```
SteRiccio, line 131: what about jszip? it's required in the code but I don't see it here...

webapp/components/Map/ShowZoomLevel.js
ramirobg94: https://www.pluralsight.com/guides/event-listeners-in-react-components

ramirobg94: You can useMapEvents or useMapEvent, they take the responsibility of mounting and unmounting the listeners

Also as you can see they unmount the listener with 'map.off' when the component is unmounted.
 https://github.com/PaulLeCam/react-leaflet/blob/1b6846de786c19ce70d8514c174b9bd020fd647f/packages/react-leaflet/src/hooks.ts#L15

webapp/components/Map/leaflet-side-by-side/index.js
```diff
@@ -0,0 +1,190 @@
+var L = require('leaflet')
+require('./layout.css')
+require('./range.css')
+
```
ramirobg94, line 4: https://www.npmjs.com/package/leaflet-side-by-side

ramirobg94: could you use this package?

webapp/views/App/views/Data/MapView/CoordinateAttributeDataLayer/CoordinateAttributePolygon.js
```diff
@@ -0,0 +1,245 @@
...
+  const { latitude, longitude } = props
+
+  const len_lat_meters = Number(polygonProperties.length_latitude) || 0
+  const len_lng_meters = Number(polygonProperties.length_longitude) || 0
+  const offset_north_meters = Number(polygonProperties.offset_north) || 0
+  const offset_east_meters = Number(polygonProperties.offset_east) || 0
+  const numberOfPointsNorth = Number(polygonProperties.number_of_points_north) || 0
+  const numberOfPointsEast = Number(polygonProperties.number_of_points_east) || 0
+  const numberOfPointsCircle = Number(polygonProperties.number_of_points_circle) || 0
+  const controlPointOffsetNorh = Number(polygonProperties.controlpoint_offset_north) || 0
```
ramirobg94, line 27: controlpointOffsetNorth or control_point_offset_north?

SteRiccio: a simple question: since the polygon properties are internal to Arena, can the name of the polygon properties be in camel case? and is it possible to use a module with getters and updaters to access and modify those properties?

webapp/components/Map/L.KML.js
Ukonhattu: That is literally the same file

webapp/components/Map/leaflet-side-by-side/index.js
Ukonhattu: I am using it, but I had to edit it because that package does not work with newer leaflet versions.

webapp/components/Map/MapBaseLayerPeriodSelector.js
```diff
@@ -50,44 +106,74 @@ export const MapBaseLayerPeriodSelector = () => {
...
+      newProc = !leftChecked ? Proc.cir : Proc.rgb
+      setState((statePev) => ({ ...statePev, leftChecked: !statePev.leftChecked, procLeft: newProc }))
+    } else {
+      newProc = !rightChecked ? Proc.cir : Proc.rgb
+      setState((statePev) => ({ ...statePev, rightChecked: !statePev.rightChecked, procRight: newProc }))
+    }
+    let layer
+    if (isLeft) {
+      layer = getLayer(true)
+    } else {
```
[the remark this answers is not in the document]
Ukonhattu: oh yeah, missed that when made it a function

package.json
Ukonhattu: hmm maybe the latest merge removed it, will check

webapp/components/Map/MapBaseLayerPeriodSelector.js
Ukonhattu: Oh, true. Just put it there quickly since it will be removed for the actual publish anyway.

webapp/components/Map/MapBaseLayerPeriodSelector.js
```diff
@@ -74,14 +74,13 @@ export const MapBaseLayerPeriodSelector = () => {
...
   const initSideBySide = () => {
-    if (User.isSystemAdmin(user)) {
-      const layerLeft = getLayer(true)
-      const layerRight = getLayer(false)
-      if (layerLeft && layerRight) {
-        sideBySideObject = L.control.sideBySide(layerLeft, layerRight).addTo(map)
-        const url = baseLayerUrlByProviderFunction[provider]({ selectedPeriodValueRight, apiKey })
-        onBaseLayerUpdate({ ...contextBaseLayer, url })
-      }
+    if (User.isSystemAdmin(user)) return
```
SteRiccio, line 77: should it be the opposite? !User.isSystemAdmin(user)?

webapp/views/App/views/Data/MapView/CoordinateAttributeDataLayer/CoordinateAttributePolygon.js
```diff
@@ -9,24 +9,24 @@ import PropTypes from 'prop-types'
 
 export const CoordinateAttributePolygon = (props) => {
   const surveyInfo = useSurveyInfo()
-  const polygonProperties = JSON.parse(surveyInfo.props.samplingPolygon)
+  const polygonProperties = surveyInfo.props.samplingPolygon || {}
```
SteRiccio, line 12: there is a getter in Survey for this, right? Survey.getSamplingPolygon, am I wrong?

webapp/components/Map/MapBaseLayerPeriodSelector.js
Ukonhattu: Yeah this we thew bug on the demo =P

Ukonhattu:
I updated leaflet-side-by-side and it now used as a dependency

SteRiccio (commented):
ok for me

ramirobg94 (approved):

PR #2395, opened 2022-09-09
Bugfix / records file import

SteRiccio:


server/modules/arenaImport/service/arenaImport/jobs/recordsImportJob.js
```diff
@@ -70,6 +70,12 @@ export default class RecordsImportJob extends Job {
       }
 
       const recordToStore = this.prepareRecordToStore(record)
+
+      const existingRecord = await RecordManager.fetchRecordByUuid(surveyId, recordUuid, this.tx)
```
ramirobg94, line 74: should we check the updateDate?

ramirobg94 (commented):
LGTM

server/modules/arenaImport/service/arenaImport/jobs/recordsImportJob.js
SteRiccio: yes, and it should be checked in the mobile app too, to avoid sending data that won't be imported

core/dateUtils.js
```diff
@@ -108,10 +108,22 @@ export const isValidDateInFormat = (dateStr, format) => {
   return fnsIsValid(parsed)
 }
 
+const substractTimezoneOffset = (date) => {
+  const timezoneOffset = date.getTimezoneOffset() * 60000
```
ramirobg94, line 112: This and 117 could be extracted to use the 60000 constant in only one place

core/dateUtils.js
```diff
@@ -108,10 +108,22 @@ export const isValidDateInFormat = (dateStr, format) => {
...
+const addTimezoneOffset = (date) => {
+  const timezoneOffset = date.getTimezoneOffset() * 60000
+  return new Date(date.getTime() + timezoneOffset)
+}
+
 export const formatDateISO = (date) => (date ? format(date, formats.dateISO) : null)
 
 export const formatDateTimeDefault = (date) => (date ? format(date, formats.datetimeDefault) : null)
 
+export const formatDateTimeISO = (date) => (date ? format(addTimezoneOffset(date), formats.datetimeISO) : null)
```
ramirobg94, line 125: The validation of date null could be extracted to the format function

server/modules/arenaImport/service/arenaImport/jobs/recordsImportJob.js
```diff
@@ -70,20 +71,47 @@ export default class RecordsImportJob extends Job {
...
+      this.incrementProcessedItems()
+    })
+
+    await nodesBatchPersister.flush()
+  }
+
+  async fetchRecordSummary({ recordUuid }) {
+    const { surveyId } = this.context
+    const { list } = await RecordManager.fetchRecordsSummaryBySurveyId({ surveyId, recordUuid }, this.tx)
+    return list[0]
```
ramirobg94, line 86: Why [0]

server/modules/arenaImport/service/arenaImport/jobs/recordsImportJob.js
```diff
@@ -70,20 +71,47 @@ export default class RecordsImportJob extends Job {
...
+
+  async insertOrSkipRecord({ record, nodesBatchPersister }) {
+    const { survey, surveyId } = this.context
+
+    const recordUuid = Record.getUuid(record)
+
+    const existingRecord = await this.fetchRecordSummary({ surveyId, recordUuid })
+
+    if (!existingRecord || DateUtils.isAfter(Record.getDateModified(record), Record.getDateModified(existingRecord))) {
+      if (existingRecord) {
```
ramirobg94, line 97: If(!existing){if (existing) ...

ramirobg94 (commented):
Some small changes

server/modules/arenaImport/service/arenaImport/jobs/recordsImportJob.js
SteRiccio: because we need to fetch only one record (summary) in this case. I don't know if it's better to have a function in RecordManager directly to do that, because it's like a fetchRecordSummaryByUuid

server/modules/arenaImport/service/arenaImport/jobs/recordsImportJob.js
SteRiccio: eheheh, it is like "insert the record if"
1) the record does not exist already
2) or the record being imported has more recent changes
in one of these 2 cases we delete the existing record only if it exists

core/dateUtils.js
SteRiccio: right again!

core/dateUtils.js
SteRiccio: right

ramirobg94 (approved):

PR #2449, opened 2022-10-04
Hide not applicable pages

### LINKED ISSUE #2446: hide not relevant entities or pages without any relevant node

SteRiccio:
resolves #2446


ramirobg94:
I think that this pages should appear because if not the user can feel confused

SteRiccio:
> I think that this pages should appear because if not the user can feel confused

there is an option in the node definition editor, "Hidden when not relevant", you can decide if the page or its descendants can be hidden or not when not relevant

ramirobg94 (approved):

PR #2462, opened 2022-10-07
Chain / reporting data categories: add cumulative area column

SteRiccio:


server/modules/category/service/categoryService.js
```diff
@@ -112,6 +112,86 @@ export const fetchSamplingPointData = async ({ surveyId, levelIndex = 0, limit,
...
+      {}
+    ),
+    ...(Category.isReportingData(category)
+      ? {
+          area_cumulative: item.areaCumulative,
+        }
+      : {}),
+  }))
+}
+
```
ramirobg94, line 194: could you use extraction to simplify this code? it is hard to follow

ramirobg94 (changes requested):

server/modules/category/service/categoryService.js
SteRiccio: of course, everything is possible

SteRiccio: moved to a separate file, a bit better but still "complex"...

PR #2587, opened 2022-12-05
Map: set initial zoom level to 3

### LINKED ISSUE #2585: Map: Open map view at the zoom level 3

SteRiccio:
fixes #2585

### DIFF
--- a/webapp/components/Map/Map.js
+++ b/webapp/components/Map/Map.js
@@ -31,6 +31,8 @@ L.Marker.prototype.options.icon = L.icon({
 })
 // end of workaround
 
+const initialZoomLevel = 3
+
 export const Map = (props) => {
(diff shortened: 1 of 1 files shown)


webapp/components/Map/Map.js
```diff
@@ -31,6 +31,8 @@ L.Marker.prototype.options.icon = L.icon({
 })
 // end of workaround
 
+const initialZoomLevel = 3
```
ramirobg94, line 34: as it is a const of configuration, maybe it is better to put upperCase

```suggestion
const INITIAL_ZOOM_LEVEL = 3
```

SteRiccio: I had the same doubt .. I agree with you, I'll change it

ramirobg94 (approved):

PR #2588, opened 2022-12-06
Survey dropdown: include survey name; allow markdown in dropdown item

### LINKED ISSUE #2582: Clone, Clone from: Show a survey name in brackets
Production contain multiple cloned surveys, created by cloning from templates. All these may contain the same default texts in survey label.
Add also the survey name into the list.

SteRiccio:
resolves #2582

### DIFF
--- a/webapp/components/survey/SurveyDropdown/useSurveyDropdownOptions.js
+++ b/webapp/components/survey/SurveyDropdown/useSurveyDropdownOptions.js
@@ -5,11 +5,26 @@ import * as Survey from '@core/survey/survey'
 import * as API from '@webapp/service/api'
 import { useI18n } from '@webapp/store/system'
 
+const getSurveyLabel = (surveyInfo) => {
+  // label will be: survey_name - survey_label
+
+  // survey name in bold (markdown)
+  const parts = [`**${Survey.getName(surveyInfo)}**`]
+
+  // survey label in survey default language
+  const label = Survey.getDefaultLabel(surveyInfo)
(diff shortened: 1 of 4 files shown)


webapp/components/survey/SurveyDropdown/useSurveyDropdownOptions.js
```diff
@@ -5,11 +5,26 @@ import * as Survey from '@core/survey/survey'
 import * as API from '@webapp/service/api'
 import { useI18n } from '@webapp/store/system'
 
+const getSurveyLabel = (surveyInfo) => {
+  // label will be: survey_name - survey_label
+
+  // survey name in bold (markdown)
+  const parts = [`**${Survey.getName(surveyInfo)}**`]
```
ramirobg94, line 12: what? now the survey label could be written with markdown?

ramirobg94: what is the reason behind? I saw the ticket, is that a requirement?

SteRiccio: The survey label cannot be written in markdown.
The requirement is to have the possibility to format somehow the label in the surveys dropdown.
In this way, when the parameter passed to the dropdown markdownInLabels is true, the text in label and description will be considered as markdown syntax.
I think it could be useful even in other cases, like variables dropdown in the expression editor. The alternative is to create a custom component for the option, but this way is easier...

PR #2929, opened 2023-07-19
Enhancement / coordinate additional fields

SteRiccio:


SteRiccio:
> shall we include changes on arena-core?

Yes, of course

PR #3040, opened 2023-09-29
Fix: now() function timezone offset

### LINKED ISSUE #3035: now() : it gives the time of the server! adjust it with user timezone

SteRiccio:
fixes #3035


SteRiccio:
> Are there any way to solve this without use the timezone? I mean using the iso date with the timezone.

I think that the easiest thing to solve this without passing the timezone or the timezone offset to the server is to store the time as a Date in ISO format, then format it in the client using its timezone... is this what you mean?

SteRiccio:
> yes, use the iso data all the time

I've just tried, the problem is that this approch introduces other pains elsewhere, e.g. in expressions that compare date or time attributes... even in that case you will have to pass the client timezone. a this point it's better to keep the storage format as it is and calculate the offset when the value is stored.

SteRiccio:
> Where are the expressions evaluated? at the same timezone of the user (client side)?

for the expressions validation client side, otherwise server side

SteRiccio:
> Where are the expressions evaluated? at the same timezone of the user (client side)?

they are evaluated in NodeDefValidator, at the same timezone of the client, but the record updater runs server side, ignoring the client timezone

